PR #3, opened 2021-09-02
various tweaks on landing page

mklgallegos:
hopefully this works


components/Features/FeaturesItem/FeaturesItem.module.css
```diff
@@ -1,8 +1,16 @@
...
+    font-weight: bold;
+    padding-bottom: 20px;
+}
+
 .container {
     padding: 24px;
     border-radius: 10px;
     border: 1px solid currentColor;
     height: 250px;
+    width: 250px;
```
seymourisdead, line 13: if you want them square, change their height to 261px and don't set the width. there are some hacks to do square blocks without setting their size in px, but since our website is not adaptive it's ok just hardcode the height.

seymourisdead:
pls run `yarn lint` and fix the errors

PR #9, opened 2022-02-07
feat: add page Team

SilencerWeb:


src/components/pages/team/cta/cta.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+import Button from 'components/shared/button';
+import Container from 'components/shared/container';
+
+const CTA = () => (
+  <section className="safe-paddings mt-48 bg-black py-36 text-center 3xl:mt-44 2xl:mt-40 2xl:py-32 xl:mt-32 xl:py-24 lg:mt-24 lg:py-20 md:mt-20">
+    <Container size="sm">
+      <h2 className="t-5xl font-bold text-white">Want to join Zenith’s team?</h2>
```
bolotskydev, line 9: So, are we alienating `Heading` component altogether? 🤔  Looks like we are only using it with `h1` tag?

bolotskydev (commented):
Perfect, nothing to add from my side 🥇

src/components/pages/team/cta/cta.jsx
SilencerWeb: Yeah but designers decided to add tons of new headings so it's not a good option to add ALL of them to `Heading`. Also, we will remove `Heading` component in the future

PR #10, opened 2022-02-08
feat: add Jobs page

SilencerWeb:


src/components/pages/jobs/hero/hero.jsx
```diff
@@ -0,0 +1,31 @@
...
+        of databases.
+      </p>
+      <div className="-mt-16 translate-y-32 lg:-mt-12 lg:translate-y-24 md:-mt-8 md:flex md:translate-y-16 md:justify-center">
+        <StaticImage
+          className="rounded-md md:min-w-[600px]"
+          src="./images/illustration.jpg"
+          alt=""
+          imgClassName="rounded-md"
+          aria-hidden
+        />
```
bolotskydev, line 25: Looks like we are missing `loading="eager"` att

src/components/pages/jobs/jobs-list/jobs-list.jsx
```diff
@@ -0,0 +1,74 @@
...
+          items: jobsByDepartmentAsObject[department],
+        }));
+
+        setItems(jobsByDepartmentAsArray);
+      })
+      .catch((error) => console.log(error));
+  }, []);
+
+  return (
+    <section className="safe-paddings pt-64 3xl:pt-60 2xl:pt-52 xl:pt-48 lg:pt-40 md:pt-28">
```
bolotskydev, line 37: I think it is better to keep spacing consistent, thus adjust a bit top padding of the job openings

image

bolotskydev (approved):
Overall looks great! Left just a few comments

src/components/pages/jobs/hero/hero.jsx
SilencerWeb: Thank you!

src/components/pages/jobs/jobs-list/jobs-list.jsx
SilencerWeb: I've made it a bit larger but not the same though since it was too much in my opinion

We'll see what designers say on review

PR #17, opened 2022-02-11
refactor: add constants file with menu links

k8isdead:
This PR brings a `constants/menus.js` file to keep all the links for the header, footer and mobile menus.


lnikell:
LGTM, later on I would suggest export `links` too, since it might be useful to use it within the page content links.

PR #21, opened 2022-02-23
fix: iterate on texts

kelvich:
Update some text per discussion in https://www.notion.so/zenithtech/Landing-texts-48ca0a6d50604149822acc164f76d530

Left hero text as it was before, since proposed rearrangement interferes with orbiting logos.


stepashka:
We did some more iterations on text, and here's the ones that are still pending that I need your help with, @lnikell 

- [ ] remove the 'BLACK FRIDAY Started' from the 'On Demand Scalability' 
- [ ] remove the 'Save 70%' from the 'Pay for the actual usage' image
- [ ] hide the 'Same PostgreSQL command as you used to will get you a smooth database creation experience' from the page for now
- [ ] could you check the headline with the blinking words? we did some experiments and made it considerably shorter. I don't have the issues with the layout, but there's a report from Stas here and Nikita complained about the same. could be only affecting some resolutions.
- [ ] 'Not an ordinary PostgreSQL as a service' has 5 items instead of 6 right now, and there's a gap for the missing one, could you help fix that?

Could you pick this up, @lnikell ?
Once these are done, you've got a green light to launch this on zenith.tech

We'll keep iterating and experimenting with the content, adding more detailed articles and documentation in the upcoming weeks.

lnikell:
@stepashka all clear, we will have to discuss the section with icons, since the way to fill the gap is bringing another feature, even if it's coming soon.

we will go over changes and let you know once it's finished.

PR #54, opened 2022-06-01
Docs cleanup quickstart

alexandriastech:


content/docs/quickstart/sqlalchemy.md
```diff
@@ -21,21 +31,24 @@ CONNSTR = f'postgresql://{USERNAME}:{TOKEN}@{PROJECT_ID}.cloud.neon.tech/main'
...
 ```
 
-References:
+## References
 
 - [Establishing Connectivity - the Engine](https://docs.sqlalchemy.org/en/14/tutorial/engine.html)
 - [Connecting to PostgreSQL with SQLAlchemy](https://docs.sqlalchemy.org/en/14/core/engines.html#postgresql)
 
-### Using from python + psycopg2
+## Step 2 — Using from Python + psycopg2
```
yeputons, line 39: I think it's not actually the second step, but a separate way of using Neon altogether, which should live in a separate file. Sqlalchemy should work as above.

alexandriastech: i have removed the numbered steps and left headings level 2 instead

alexandriastech:
> Is it intended that this PR also includes the Overview section?
> 
> I suspect this is due to the branch `docs-cleanup-quickstart` starting from the `docs-cleanup` branch, but submitted into `main`. I would suggest to start the branches from the same branch they are going to be merged into, or to be more specific, make sure each change is part of just one PR, not multiple.
> 
> Shall we redirect this PR to docs-cleanup?


Hi @stepashka 
- This PR is dedicated to the Quickstart section only (I have 3 early commits that should have been in the docs-cleanup only).
- My idea is to close both docs-cleanup and docs-cleanup-quickstart and start over. 
Would you like us still to dedicate a branch to each section of the Neon docs (which was my assumption from our Tuesday meeting) or work all in one main PR?

PR #64, opened 2022-06-08
change information architecture of neon

alexandriastech:


alexandriastech:
@SilencerWeb Hello there! We are looking to see if it is possible to add another nested level in the left navigational sidebar, do you know how we can achieve this?

Thank you in advance!

PR #75, opened 2022-06-17
fix: add custom event on success submit early access form

seymourisdead:
not sure if i did it right


SilencerWeb (approved):
Code seems ok to me

PR #91, opened 2022-07-15
Adds a landing page for links from marketing emails

WeaponXII:
- This PR adds a landing page that leads to a list of Neon's integrations
- This page is not accessible from the sidebar, this would be an anti-pattern because it has the same functionality as the sidebar list of integrations
- This page is only accessible from the direct link (and the redirected link), it has been implemented specifically so that Neon can keep directing users to a list of integrations from a specific URL
- This adds Neon integrations that were not available when the original page was created and links to each integration in the page for user convenience
- It also redirects from the much older deprecated integrations URL slug so that it receives any traffic that link may still be attracting


klink (dismissed):

content/docs/integrations/integrations-list.md
```diff
@@ -1,22 +1,25 @@
 ---
 title: Integrations
+redirectFrom:
+  - docs/cloud/integrations
 ---
 
-To see **Cloud Integrations**,
+To see **Cloud Integrations** that Neon integrates with,
```
danieltprice, line 7:
```suggestion
For Neon **Cloud integrations**, see:
```

content/docs/integrations/integrations-list.md
```diff
@@ -1,22 +1,25 @@
...
 
-- Run a Hasura App
-- Run a Next.js and Vercel App
-- Run Prisma App
+- Run a [Hasura App](../hasura)
+- Run a [Next.js and Vercel App](../vercel)
+- Run a [Prisma App](../prisma)
 
 To see **Frameworks & Programming Languages** that Neon integrates with,
```
danieltprice, line 13:
```suggestion
For Neon **Framework & Programming Language** integrations, see:
```

content/docs/integrations/integrations-list.md
```diff
@@ -1,22 +1,25 @@
...
-- Run a Symphony App
+- Run a [Go App](../go)
+- Run a [Java App](../java)
+- Run a [Django App](../django)
+- Run a [Node.js App](../node)
+- Run a [Symfony App](../symfony)
+- Run a [Rust App](../rust)
 
-To see **Others**,
+To see **Other** Neon integrations,
```
danieltprice, line 22: For other Neon integrations, see:

danieltprice, line 22:
```suggestion
For other Neon integrations, see:
```

content/docs/integrations/integrations-list.md
```diff
@@ -1,22 +1,25 @@
 ---
 title: Integrations
+redirectFrom:
+  - docs/cloud/integrations
 ---
```
danieltprice, line 6: Can we add a short introductory sentence, such as:
"This page lists the available Neon integrations."

danieltprice (commented):
A few suggested revisions. If possible, I'd like to see a build of this page before it's merged. I am not sure about the bold items within sentences. Perhaps these would be better as subheadings:
## Cloud integrations
## Frameworks & programming languages
## Other integrations

danieltprice:
Agree that this page should not be introduced as is. If we need a single landing page for integrations, let's revisit when that need arises again.

PR #104, opened 2022-07-29
Release notes for 2022-07-20

duskpoet:
This PR adds the Release Notes for the [2022-07-20 release of the Console](https://github.com/neondatabase/cloud/pull/1858/commits)

### DIFF
--- a/content/docs/release-notes.md
+++ b/content/docs/release-notes.md
@@ -6,7 +6,7 @@ title: Release Notes
 
 ### What's new
 
-* UI for pool enabling
+* UI: add 'Enable pooling' toggle to the Project's Settings page 
 
 ## 2022-07-19 Storage release
 
--- a/content/docs/release-notes.md
+++ b/content/docs/release-notes.md
@@ -6,7 +6,9 @@ title: Release Notes
 
 ### What's new
 
-* UI: add 'Enable pooling' toggle to the Project's Settings page 
+* UI: add 'Enable pooling' toggle to the Project's Settings page.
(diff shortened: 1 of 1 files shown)


content/docs/release-notes.md
```diff
@@ -2,6 +2,12 @@
 title: Release Notes
 ---
 
+## 2022-07-20 Console release
+
+### What's new
+
+* UI for pool enabling
```
stepashka, line 9:
```suggestion
* UI: add 'Enable pooling' toggle to the Project's Settings page 
```

stepashka (commented):
Have we added more instances to the console as part of this release? This enables the 0 downtime deployment, i think we can mention that too.
We're also more precise when specifying the Postgres version, we can mention that too.
Even though this isn't published yet, we can mention that work has started on the api v2 and the spec is available now.

Anything else worth mentioning? Will add other reviewers here too.

duskpoet:
> Have we added more instances to the console as part of this release? This enables the 0 downtime deployment, i think we can mention that too. We're also more precise when specifying the Postgres version, we can mention that too. Even though this isn't published yet, we can mention that work has started on the api v2 and the spec is available now.
> 
> Anything else worth mentioning? Will add other reviewers here too.

Unfortunately I am not really familiar with other changes. Maybe @ololobus @antonyc can help

ololobus (approved):
Pushed more details. I suggest to avoid mentioning v2 now

stepashka (approved):

PR #113, opened 2022-08-09
docs: add `enableTableOfContents` field to readme

stepashka:
Add documentation about how to use the enableTableOfContents field.

### DIFF
--- a/content/docs/README.md
+++ b/content/docs/README.md
@@ -39,6 +39,7 @@ Right now Markdown files accept 2 fields:
 1. `title` — title of the page (required)
 2. `redirectFrom` — array of strings with paths to redirect from to the page, should start and end with a slash, e.g. `/docs/old-path/`
 3. `isDraft` — flag that says the page is not ready yet. It won't appear in production but will appear in the development mode.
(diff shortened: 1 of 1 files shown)


stepashka:
I've added a mention of the `enableTableOfContents` to the docs readme.
still missing description about which headers will be included in the outline/TOC
e.g. would TOC include only the top level headers, only the first or the second level, or something else?
could you please add that?
thank you!

content/docs/README.md
```diff
@@ -39,6 +39,7 @@ Right now Markdown files accept 2 fields:
 1. `title` — title of the page (required)
 2. `redirectFrom` — array of strings with paths to redirect from to the page, should start and end with a slash, e.g. `/docs/old-path/`
 3. `isDraft` — flag that says the page is not ready yet. It won't appear in production but will appear in the development mode.
+4. `enableTableOfContents` — flag that turns on the display of the outline for the page.
```
bolotskydev, line 42:
```suggestion
4. `enableTableOfContents` — flag that turns on the display of the outline for the page. The outline gets built out of second and third-level headings ([`h2`, `h3`]), thus appears as two-level nested max.
```

bolotskydev (approved):

PR #130, opened 2022-08-26
feat: add 404 not found page

vannyle:
This pull request introduces the Not found page - 404

**Step by steps**
1. Open [page](https://websitemain62807-notfound.gtsb.io/404/) 
2. Check if everything works as expected


bolotskydev (dismissed):

stepashka (commented):
> Sorry, the page you are looking for doesn’t exist or has been moved. We suggest you back to home.

this sounds odd.
1) we already have a green button CTA 'Back Home' no need to repeat that in the error text, please remove the 'We suggest you back to home.'
2) can we also add the docs search box on the 404 page?

lnikell (approved):

PR #139, opened 2022-08-31
fix: update SNI doc

kelvich:
Be more strict with project id/name difference. In passing fix awkward
wording and update status for golang drivers.

### DIFF
--- a/content/docs/how-to-guides/connectivity-issues.md
+++ b/content/docs/how-to-guides/connectivity-issues.md
@@ -7,7 +7,7 @@ title: Connecting with old clients
 In most cases, copy-pasting `Connection string` from the project's dashboard and using it in your project should work as is. However, with older clients and some native Postgres clients, you may receive the following error:
 
 ```txt
-ERROR: Project ID is not specified. Either please upgrade the postgres client library (libpq) for SNI support or pass the project ID (first part of the domain name) as a parameter: '&options=project%3D<project-id>'. See more at https://neon.tech/sni
+ERROR: The project ID is not specified. Either upgrade the PostgreSQL client library (libpq) for SNI support or pass the project ID (the first part of the domain name) as a parameter: '&options=project%3D'. See https://neon.tech/sni for more information.
 ```
 
 In most cases, this happens if your client library or app does not support the so-called **SNI (Server Name Indication)** mechanism in TLS. See [#Details](Details) for more context and [#Workarounds](Workarounds) for a list of ways to get around this issue.
--- a/content/docs/how-to-guides/connectivity-issues.md
+++ b/content/docs/how-to-guides/connectivity-issues.md
@@ -56,7 +56,7 @@ If your application or service uses golang Postgres clients like `pgx` and `lib/
 
 ### D. Specify the project ID in the password field
 
-You can specify the project ID in the password field as a last resort. So instead of your actual password (let it be 'my-pass'), you can set the following string, consisting of the project ID and the password:
+You can specify the project ID in the password field as a last resort. So, instead of your actual password (for example, 'my-pass'), you can set the following string, consisting of the project ID and the password:
 
 ```txt
 project=mute-recipe-239816;my-pass
--- a/content/docs/how-to-guides/connectivity-issues.md
+++ b/content/docs/how-to-guides/connectivity-issues.md
@@ -7,20 +7,20 @@ title: Connecting with old clients
 In most cases, copy-pasting `Connection string` from the project's dashboard and using it in your project should work as is. However, with older clients and some native Postgres clients, you may receive the following error:
 
 ```txt
-ERROR: Project name is not specified. Either please upgrade the postgres client library (libpq) for SNI support or pass the project name as a parameter: '&options=project%3D<project-name>'. See more at https://neon.tech/sni
+ERROR: Project ID is not specified. Either please upgrade the postgres client library (libpq) for SNI support or pass the project ID (first part of the domain name) as a parameter: '&options=project%3D<project-id>'. See more at https://neon.tech/sni
 ```
 
 In most cases, this happens if your client library or app does not support the so-called **SNI (Server Name Indication)** mechanism in TLS. See [#Details](Details) for more context and [#Workarounds](Workarounds) for a list of ways to get around this issue.
 
 ## Details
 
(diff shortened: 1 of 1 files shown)


content/docs/how-to-guides/connectivity-issues.md
```diff
@@ -7,20 +7,20 @@ title: Connecting with old clients
 In most cases, copy-pasting `Connection string` from the project's dashboard and using it in your project should work as is. However, with older clients and some native Postgres clients, you may receive the following error:
 
 ```txt
-ERROR: Project name is not specified. Either please upgrade the postgres client library (libpq) for SNI support or pass the project name as a parameter: '&options=project%3D<project-name>'. See more at https://neon.tech/sni
+ERROR: Project ID is not specified. Either please upgrade the postgres client library (libpq) for SNI support or pass the project ID (first part of the domain name) as a parameter: '&options=project%3D<project-id>'. See more at https://neon.tech/sni
```
danieltprice, line 10: ERROR: The project ID is not specified. Either upgrade the PostgreSQL client library (libpq) for SNI support or pass the project ID (the first part of the domain name) as a parameter: '&options=project%3D<project-id>'. See https://neon.tech/sni for more information.

content/docs/how-to-guides/connectivity-issues.md
```diff
@@ -7,20 +7,20 @@ title: Connecting with old clients
...
+To route incoming connections, we use different domain names for different projects, e.g., to connect to the project `mute-recipe-239816`, we ask you to connect to _**mute-recipe-239816.cloud.neon.tech**_. However, the Postgres wire protocol does not transfer the server domain name, so we rely on the so-called **SNI (Server Name Indication)** extension of the `TLS` protocol, which allows a client to indicate what domain name it is attempting to connect to. That is the same mechanism that allows hosting several `https`-enabled websites on a single IP address. `SNI` support was added to the `libpq` (an official Postgres client library) in version 14, released in September 2021. All `libpq`-based clients like Python's `psycopg2` and Ruby's `ruby-pg` should work if `libpq` in the system has a version >= 14.
```
danieltprice, line 17: To route incoming connections, Neon uses different domain names for different projects; for example, to connect to the project `mute-recipe-239816`, we ask you to connect to _**mute-recipe-239816.cloud.neon.tech**_. However, the PostgreSQL wire protocol does not transfer the server domain name, so Neon relies on the **SNI (Server Name Indication)** extension of the `TLS` protocol, which allows a client to indicate the domain name it is attempting to connect to. This is the same mechanism that allows hosting several `https`-enabled websites on a single IP address. `SNI` support was added to `libpq` (an official PostgreSQL client library) in version 14, released in September, 2021. All `libpq`-based clients like Python's `psycopg2` and Ruby's `ruby-pg` should work if the `libpq` library on the system is version 14 or higher.

content/docs/how-to-guides/connectivity-issues.md
```diff
@@ -7,20 +7,20 @@ title: Connecting with old clients
...
 
 ## Workarounds
 
-If you face a `Project name is not specified` error and library/application upgrade does not help, we provide several fallback options. When `SNI` domain name info is missing, we need to get it via other means.
+If you face a `Project ID is not specified` error and library/application upgrade does not help, we provide several fallback options. When `SNI` domain name info is missing, we need to get it via other means.
```
danieltprice, line 21: If you encounter a `Project ID is not specified` error and a library or application upgrade does not help, we provide several fallback options. When `SNI` domain name information is missing, we need to obtain this information through other means.

content/docs/how-to-guides/connectivity-issues.md
```diff
@@ -54,9 +54,9 @@ This option is expected to work with all libpq-based apps.
 
 If your application or service uses golang Postgres clients like `pgx` and `lib/pg` you can set `sslmode=verify-full,` which will cause `SNI` info to be sent. Most likely, this was not intentional but happened inadvertently due to the golang's TLS library API design.
 
-### D. Specify the project name in the password field
+### D. Specify the project ID in the password field
 
-You can specify the project name in the password field as a last resort. So instead of your actual password (let it be 'my-pass'), you can set the following string, consisting of the project name and the password:
+You can specify the project ID in the password field as a last resort. So instead of your actual password (let it be 'my-pass'), you can set the following string, consisting of the project ID and the password:
```
danieltprice, line 59: You can specify the project ID in the password field as a last resort. So, instead of your actual password (for example,  'my-pass'), you can set the following string, consisting of the project ID and the password:

content/docs/how-to-guides/connectivity-issues.md
stepashka, line 10:
```suggestion
ERROR: The project ID is not specified. Either upgrade the PostgreSQL client library (libpq) for SNI support or pass the project ID (the first part of the domain name) as a parameter: '&options=project%3D'. See https://neon.tech/sni for more information.
```

content/docs/how-to-guides/connectivity-issues.md
danieltprice, line 17:
```suggestion
To route incoming connections, we use different domain names for different projects, e.g., to connect to the project `mute-recipe-239816`, we ask you to connect to _**mute-recipe-239816.cloud.neon.tech**_. However, the Postgres wire protocol does not transfer the server domain name, so we rely on the so-called **SNI (Server Name Indication)** extension of the `TLS` protocol, which allows a client to indicate what domain name it is attempting to connect to. That is the same mechanism that allows hosting several `https`-enabled websites on a single IP address. `SNI` support was added to the `libpq` (an official Postgres client library) in version 14, released in September 2021. All `libpq`-based clients like Python's `psycopg2` and Ruby's `ruby-pg` should work if `libpq` in the system has a version >= 14.
```

content/docs/how-to-guides/connectivity-issues.md
danieltprice, line 21:
```suggestion
If you encounter a Project ID is not specified error and a library or application upgrade does not help, we provide several fallback options. When SNI domain name information is missing, we need to obtain this information through other means.
```

content/docs/how-to-guides/connectivity-issues.md
danieltprice, line 59:
```suggestion
You can specify the project ID in the password field as a last resort. So instead of your actual password (let it be 'my-pass'), you can set the following string, consisting of the project ID and the password:
```

danieltprice, line 59:
```suggestion
You can specify the project ID in the password field as a last resort. So, instead of your actual password (for example, 'my-pass'), you can set the following string, consisting of the project ID and the password:
```

danieltprice (approved):

kelvich:
@danieltprice, thanks for fixing it!

PR #140, opened 2022-09-01
nit: fix doc

nikitakalyanov:
I think there should be the 'cold' word here. Moreover, we already have 'cold' here in a similar context: https://github.com/neondatabase/website/blob/897b3bda15dc96b0a90858e9354a6516a3eea042/src/components/pages/home/storage/storage.jsx#L29

### DIFF
--- a/src/components/pages/home/advantages/advantages.jsx
+++ b/src/components/pages/home/advantages/advantages.jsx
@@ -35,7 +35,7 @@ const Advantages = () => {
       animationData: costEfficientAnimationData,
       title: 'Cost Efficient',
       description:
-        'Compute scales down to zero on usage and hot storage offloads to S3 for cost efficiency.',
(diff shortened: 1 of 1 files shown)


src/components/pages/home/advantages/advantages.jsx
```diff
@@ -35,7 +35,7 @@ const Advantages = () => {
       animationData: costEfficientAnimationData,
       title: 'Cost Efficient',
       description:
-        'Compute scales down to zero on usage and hot storage offloads to S3 for cost efficiency.',
+        'Compute scales down to zero on usage and cold storage offloads to S3 for cost efficiency.',
```
danieltprice, line 38: Compute scales down to zero on usage, and cold storage offloads to S3 for cost efficiency.

nikitakalyanov: fixed

danieltprice (approved):

PR #169, opened 2022-09-14
docs: copied the branching guide into repo

WeaponXII:
This copies the existing branching guide into the repo, the only additions are the title in the frontmatter and draft has been set to true. It will be changed in a later PR into website content by removing the email template variable and adding website-specific content.

### DIFF
--- a/content/docs/how-to-guides/using-branches.md
+++ b/content/docs/how-to-guides/using-branches.md
@@ -1,0 +1,52 @@
+---
+title: How to Use Branches
+draft: true
+---
+Hey %NAME%!
+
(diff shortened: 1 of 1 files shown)


danieltprice:
I will be adding some branching content as well. Let's get this into the repo, and then we can look at what we want overall for the branching docs.

danieltprice (approved):

WeaponXII:
Got it @danieltprice , for some added context on this I will be rewriting this to make it into an article (was formerly an email template) and expanding this page some more into a how-to guide in a future PR

PR #173, opened 2022-09-15
fix: add padding left for ol in content

vannyle:
This pull request adds padding-left to an ordered list for content block.
Preview: https://websitemain62807-contentstyles.gtsb.io/docs/get-started-with-neon/query-with-neon-sql-editor/


danieltprice (approved):
Thank you, Vanny. That looks good to me!

PR #188, opened 2022-09-23
Fix naming inconsistencies

danieltprice:
- Postgres > PostgreSQL (Heikki and Stas have agreed to use the official name "PostgreSQL" in the docs)
- Control plane > Control Plane (Release notes)
- A few other minor edits


stepashka:
what's the thinking behind the ~~Postgres~~ Compute, @danieltprice ?

stepashka (approved):
@danieltprice understand the Postgres -> PostgreSQL, but i wonder why not PostgreSQL Compute?
otherwise approved

danieltprice:
> @danieltprice understand the Postgres -> PostgreSQL, but i wonder why not PostgreSQL Compute?

We had both Compute and PostgreSQL Compute. If we feel that the latter is more accurate, we can certainly go with that.

Issue #180, opened 2022-09-19
Document Neon domains


danieltprice:
pg.neon.tech (for link auth)
connect.neon.tech (also for link auth)
project_id.cloud.neon.tech (old but still supported domain)
endpoint_id.region_slug.platform.neon.tech (domain for SNI-enabled clients)

danieltprice:
Not seeing the need for this right now.

PR #201, opened 2022-09-30
Dprice drop postgres app topic

danieltprice:
- Remove Postgres app topic from the sidebar. It does not belong in the integrations section.
- Add relevant content to the Query with psql topic
- Adjust redirects
- https://websitemain62807-dpricedroppostgresapptopic.gtsb.io/docs/get-started-with-neon/query-with-psql-editor/


raoufchebri (approved):

content/docs/integrations/postgres.md
```diff
@@ -1,19 +1,13 @@
...
 
-Since Neon compute is regular PostgreSQL, any PostgreSQL app will run out of the box.
-You can also use all the standard client utilities like `psql` and `pg_dump`, client libraries and drivers to connect.
+A Neon compute runs regular PostgreSQL, which means that you run any PostgreSQL app with Neon.
 
-All connections to PostgreSQL are via an endpoint hosted at `pg.neon.tech`.
+You can also connect to Neon using standard PostgreSQL client utilities such as `psql` and `pg_dump`, standard client libraries, and drivers.
 
-Try connecting to `pg.neon.tech` using the following command in your terminal:
+All connections to Neon PostgreSQL are via an endpoint hosted at `pg.neon.tech`.
```
stepashka, line 9: I'm not sure that this is the only endpoint we support.
Not sure if we have the same proxy for the passwordless auth and for the normal connections.
Please double check?

danieltprice: I have set this topic to isDraft to exclude it from Algolia search, so users will not see this content. I will eventually remove the file entirely. 
I have an issue open document the endpoints that we support: #180

stepashka (approved):

PR #203, opened 2022-09-30
Update rust.md

danieltprice:
Initial Rust topic revisions.
https://websitemain62807-dpricerusttopicrevisions.gtsb.io/docs/integrations/rust/

### DIFF
--- a/content/docs/integrations/rust.md
+++ b/content/docs/integrations/rust.md
@@ -6,8 +6,6 @@ redirectFrom:
 
 This topic describes how to create a Neon project and connect to it from a Rust app.
 
-To connect to Neon from a Rust app:
-
 1. [Create a Neon project](#create-a-neon-project)
 2. [Configure the connection](#configure-the-connection)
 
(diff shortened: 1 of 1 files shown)


content/docs/integrations/rust.md
```diff
@@ -17,4 +36,11 @@ fn main() {
 }
 ```
 
-[On rust-lang playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0d9daa9cde3c74d2916c8f05b24707a3)
+where:
+
+- `<user>` is the database user, which is found on the project **Dashboard**, under **Connection Details**.
+- `<dbname>` is the database name (the default Neon project database is `main`).
+- `<password>` is the database user's password, which is provided to you when you create a Neon project.
+- `<project_id>` is the ID of the Neon project, which is found on the project **Settings** tab, under **General Settings**.
```
SomeoneToIgnore, line 44: That parameter seems to be unused in the code above?

Also `dbname` is `db_name` there.

danieltprice: Fixed. Thank you!

SomeoneToIgnore (commented):
Great, now let's update the playground link accordingly (https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0d9daa9cde3c74d2916c8f05b24707a3 now has different id names)

I wonder, whether we need such playground link at all, but have no real objections against it either.

content/docs/integrations/rust.md
danieltprice: Dropped the rust playground link. The example is now out of sync with the doc.

SomeoneToIgnore (approved):

PR #205, opened 2022-09-30
Next.js (Vercel) topic revisions.

danieltprice:
- Current Next.js (Vercel) topic revisions, We'll revisit this content for the Vercel integration.
- Move under Develop heading in TOC
https://websitemain62807-dpriceverceltopicrevisions.gtsb.io/docs/integrations/vercel/

### DIFF
--- a/content/docs/integrations/vercel.md
+++ b/content/docs/integrations/vercel.md
@@ -61,7 +61,7 @@ import postgres from 'postgres';
 
 const sql = postgres(process.env.DATABASE_URL);
 
-const result = await sql.uafe(req.body);
+const result = await sql.unsafe(req.body);
 ```
 
 _**Important**_: Never expose your Neon credentials to the browser.
--- a/content/docs/integrations/vercel.md
+++ b/content/docs/integrations/vercel.md
@@ -17,7 +17,7 @@ To create a Neon project and access it from a Next.js app:
 
 ## Create a Neon project
 
-When creating a Neon project, take note of your project ID, database name, user, password, and port number. This information is required when defining your connection settings. Neon provides the password to you immediately after you create a project.
+When creating a Neon project, take note of your project ID, database name, user, password, and port number. This information is required when defining your connection settings. 
 
 To create a Neon project:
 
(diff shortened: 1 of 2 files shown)


content/docs/integrations/vercel.md
```diff
@@ -1,43 +1,63 @@
...
+
+Add your Neon connection details to your `.env` file.
 
 ```shell
-NEON_HOST=...
-NEON_DB=...
-NEON_USER=...
-NEON_PASS=...
-NEON_PORT=...
+NEON_HOST='<project_id>.cloud.neon.tech'
```
duskpoet, line 43: I think connection string method is rather easier to use

danieltprice: Hi @duskpoet 
Happy to use the connection string instead. In this case, do I simply drop the connection string with the password included in the .env file as provided on the Dashboard?

And in that case, how should I modify the connection snippet that we provide?

import postgres from 'postgres';

const sql = postgres({
  host: process.env.NEON_HOST,
  port: process.env.NEON_PORT,
  database: process.env.NEON_DB,
  username: process.env.NEON_USER,
  password: process.env.NEON_PASS,
});

const result = await sql.uafe(req.body);

duskpoet: .env
```
DATABASE_URL=<neon_connection_string>
```
index.ts
```
const sql = postgres(process.env.DATABASE_URL);
```

danieltprice: Thank you!

content/docs/integrations/vercel.md
```diff
@@ -1,56 +1,67 @@
...
-const sql = postgres({
-  host: process.env.NEON_HOST,
-  port: process.env.NEON_PORT,
-  database: process.env.NEON_DB,
-  username: process.env.NEON_USER,
-  password: process.env.NEON_PASS,
-});
+const sql = postgres(process.env.DATABASE_URL);
 
 const result = await sql.uafe(req.body);
```
duskpoet, line 64:
```suggestion
const result = await sql.unsafe(req.body);
```

duskpoet (approved):

duskpoet (approved):

PR #214, opened 2022-10-04
Terminology fixes

danieltprice:
Use "regular PostgreSQL" instead of "vanilla PostgreSQL". We currently use both. We need to settle on one. I think "regular" is less ambiguous than "vanilla" for users who are not familiar with PostgreSQL lingo.

I will add a glossary definition for "regular PostgreSQL" in a separate PR for the glossary.

### DIFF
--- a/content/docs/conceptual-guides/architecture-overview.md
+++ b/content/docs/conceptual-guides/architecture-overview.md
@@ -6,7 +6,7 @@ redirectFrom:
 
 Neon architecture is based on the separation of storage and compute and is orchestrated by the Neon Control Plane, which manages cloud resources across both storage and compute.
 
-A Neon Compute runs regular PostgreSQL, and storage is a multi-tenant key-value store for PostgreSQL pages, which is custom-built for the cloud.
+A Neon Compute runs PostgreSQL, and storage is a multi-tenant key-value store for PostgreSQL pages that is custom-built for the cloud.
 
 ![Neon architecture diagram](/docs-images/neon_architecture_2.png)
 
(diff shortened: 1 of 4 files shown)


sharnoff:
> I think "regular" is less ambiguous than "vanilla" for users who are not familiar with PostgreSQL lingo.

That's interesting - I would have assumed "vanilla" is more clear (as someone with not much experience using postgres). Separately: I might be unusual here, but IMO "regular" feels like an overloaded term - not immediately clear that it's referring to something specific (or: referring to something specific, but as a reader I might assume I already knew what it meant). What are your thoughts on phrases like "regular/normal/typical PostgreSQL _instance_"?

Not intending to cause too much bikeshedding -- if you feel "regular PostgreSQL" is best, go for it :)

> I will add a glossary definition for "regular PostgreSQL" in a separate PR for the glossary.

Adding a glossary definition makes a lot of sense to me.

danieltprice:
> > I think "regular" is less ambiguous than "vanilla" for users who are not familiar with PostgreSQL lingo.
> 
> That's interesting - I would have assumed "vanilla" is more clear (as someone with not much experience using postgres). Separately: I might be unusual here, but IMO "regular" feels like an overloaded term - not immediately clear that it's referring to something specific (or: referring to something specific, but as a reader I might assume I already knew what it meant). What are your thoughts on phrases like "regular/normal/typical PostgreSQL _instance_"?
> 
> Not intending to cause too much bikeshedding -- if you feel "regular PostgreSQL" is best, go for it :)
> 
> > I will add a glossary definition for "regular PostgreSQL" in a separate PR for the glossary.
> 
> Adding a glossary definition makes a lot of sense to me.
Thank you for the feedback, Max. I was a bit unsure what vanilla meant when I first saw it, but that could just be me! Let me throw it out to the team for input. I a mostly concerned with having a single term.

I do like "PostgreSQL _instance_". Referring to "a Postgres" seems odd.

danieltprice:
"regular PostgreSQL" and "vanilla PostgreSQL" were dropped in favor of "standalone PostgreSQL installation".

sharnoff (approved):
Looks good! +1 for "standalone PostgreSQL installlation"

danieltprice:
Thank you, @sharnoff

PR #215, opened 2022-10-04
Dprice partner integration doc

danieltprice:
This is an initial draft of a partner integration document. It has not been validated.

### DIFF
--- a/content/docs/oauth.md
+++ b/content/docs/oauth.md
@@ -1,6 +1,7 @@
 ---
 title: Configuring Neon OAuth for 3rd party Neon API access
 enableTableOfContents: true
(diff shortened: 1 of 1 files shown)


antonyc:
Is this expected to be accessible by users publicly?

raoufchebri:
it seems we're early for this to be published

danieltprice:
Closing this pull request. Instructions can be generated from the cloud repo guide.

acervantes23:
Thanks Daniel! I think we should return to this after "GA" (I know we changed the name) since we have more urgent matters to attend to. @stepashka what do you think?

PR #241, opened 2022-10-17
Update technical-preview-free-tier.md

danieltprice:
Comment out the query that does not work:
https://websitemain62807-dpricedatasizecontentupdates.gtsb.io/docs/reference/technical-preview-free-tier/

### DIFF
--- a/content/docs/reference/technical-preview-free-tier.md
+++ b/content/docs/reference/technical-preview-free-tier.md
@@ -18,14 +18,14 @@ Technical Preview Free Tier users can create up to three projects in Neon. Proje
 Neon separates storage and compute and stores data in its own internal format.
 The 10GB data size limit applies to the logical size of a Neon project. The logical size is the sum of all relation sizes in the project.
 
-<!--
-For users familiar with PostgreSQL, the logical size is roughly equal to:
+To check the logical size of the relations in your Neon project, run the following query:
 
 ```sql
 SELECT pg_size_pretty(sum(pg_database_size(datname)))
 FROM pg_database;
 ```
 
+<!--
 To check the logical size of your Neon project, run the following query:
 
 ```sql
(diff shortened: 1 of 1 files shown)


knizhnik (commented):
Sorry, I still do not understand why do we need to comment  usage of `pg_database_size()`.
If we limit logical size of the database, then definitely users should have some way to check the current size, otherwise limit exhaustion error will be  unpleasant surprise.

Concerning `pg_cluster_size` and `neon` extension - them are needed mostly for us. Their descriptoin can really be removed from documentation.

danieltprice:
Thank you, @knizhnik. I misunderstood. I thought the first query was was also not ideal. I have added it back.

stepashka:
let's move out the details about how limita are calculated out of the free tier article
the 'Reference' section may be a good place for that

stepashka (commented):
let's move out the details about how limita are calculated out of the free tier article
the 'Reference' section may be a good place for that

danieltprice:
> let's move out the details about how limita are calculated out of the free tier article the 'Reference' section may be a good place for that

content/docs/reference/technical-preview-free-tier.md
```diff
@@ -18,18 +18,20 @@ Technical Preview Free Tier users can create up to three projects in Neon. Proje
 Neon separates storage and compute and stores data in its own internal format.
 The 10GB data size limit applies to the logical size of a Neon project. The logical size is the sum of all relation sizes in the project.
 
-For users familiar with PostgreSQL, the logical size is roughly equal to:
+To check the logical size of the relations in your Neon project, run the following query:
```
stepashka, line 21: not sure every one of our users would be familiar with the word 'relations'

stepashka (approved):
not sure every one or even most of our users would be familiar with the word 'relations'

let's fix it soon, but not insisting on delaying this merge for that.

content/docs/reference/technical-preview-free-tier.md
danieltprice: Let's change that to databases!

stepashka (approved):

PR #264, opened 2022-10-26
feat: index staging algolia queries

### LINKED ISSUE #184: Address Algolia issues
1. Clicks in Algolia search results are not being tracked properly 
2. Algolia search results include doc files not included in the TOC. Indices need to be rebuilt, but how do we exclude files from being indexed again?
3. Where can we place draft files so that they do not show up in search results? Can we exclude files or folders in the content directory from being indexed?

vannyle:
This pull request includes the draft doc posts to deploy preview, and doc posts will be indexed by Algolia queries. 

Closes #184 p.2


.env.example
```diff
@@ -3,7 +3,8 @@ GATSBY_DEV_SSR=false
 
 GATSBY_ALGOLIA_APP_ID=
 GATSBY_ALGOLIA_SEARCH_KEY=
-GATSBY_ALGOLIA_INDEX_NAME=preview_neon_docs
+GATSBY_ALGOLIA_INDEX_NAME=
+GATSBY_ALGOLIA_INDEX_NAME_STAGING=preview_neon_docs
```
bolotskydev, line 7:
```suggestion
GATSBY_ALGOLIA_INDEX_NAME_STAGING=preview_neon_docs
```

src/utils/constants.js
```diff
@@ -0,0 +1,10 @@
+const BRANCH = process.env.BRANCH || 'dev';
```
bolotskydev, line 1: It doesn't quite make sense to keep this variable of a `string` type as it is a boolean condition, it is either production or not. Besides, this variable can be confusing - it will match of course the branch name in prod, but I doubt all other branches are named `dev` 😉

bolotskydev: Let's rename it everywhere to `IS_PRODUCTION`

bolotskydev (approved):

Issue #291, opened 2022-11-07
Embeded video not showing up


danieltprice:
The videos we embed on site pages using <video></video> tags are not showing up on mobile screens.
https://neon.tech/docs/guides/hasura/
https://neon.tech/blog/neon-hasura-integration/

PR #301, opened 2022-11-08
fix: video issue on hasura page

vannyle:
This pull request update the video codec from `h264` to `h265`, as on some Android devices h264 codec is not supported https://bugs.chromium.org/p/chromium/issues/detail?id=719023

Issue: https://github.com/neondatabase/website/issues/291


danieltprice (approved):
Thank you @vannyle

PR #326, opened 2022-11-18
Update security page

zoete:
Updated to reflect that we now have regions outside of the USA, also bumped the year for the security roadmap


danieltprice (approved):
Thanks for the update, Rory. Lgtm!

PR #327, opened 2022-11-18
Update regions image

danieltprice:
The previous was blurry. Try again.


danieltprice:
Not possible to take an image 2x rendered size. Opened PP issue to find a different solution.

PR #330, opened 2022-11-18
Update prisma.md

danieltprice:
Add note about adjusting the connection string in Prisma when connection pooling is enabled.
https://websitemain62807-dpriceupdateprismadocspgbounce.gatsbyjs.io/docs/guides/prisma/


vadim2404:
Looks good. My suggestion is to cover the topic of prepared statements and connection pooling. It's not a neon issue, it's how it pgbouncer works.

PR #331, opened 2022-11-18
Add storage release notes (2022-11-16)

petuhovskiy:


content/release-notes/2022-11-16.md
```diff
@@ -0,0 +1,11 @@
+---
+label: 'Storage'
+---
+
+### What's new
+
+- Pageserver, Safekeeper, Compute, and Proxy: Reduced the size of Neon storage binaries by 50% by removing dependency debug symbols from the release build.
+- Pageserver: Moved the Write-Ahead Log (WAL) redo process code from Neon's `postgres` repository to the `neon` repository and created a separate `wal_redo` binary in order to reduce the amount of change in the `postgres` repository codebase.
+- Compute: Added prefetching support to improve sequential scan performance. Prefetching support is disabled by default but can be enabled by using `set enable_seqscan_prefetch = on; set seqscan_prefetch_buffers = 100;`.
```
danieltprice, line 9: @petuhovskiy Will end users use these variables or are they "managed" by us?

danieltprice (dismissed):
lgtm

content/release-notes/2022-11-16.md
petuhovskiy: > Will end users use these variables or are they "managed" by us?

Not sure, AFAIK currently we tell some of our users to try to enable that option to see how it improves performance, but we don't generally enable it because there might be bugs and we want to continue work on improving prefetch performance. I don't have a lot of context for this, so maybe @MMeent and @hlinnaka can give more details about what we should write in the changelog.

danieltprice (approved):
Revision.

PR #342, opened 2022-11-22
Update hero.jsx to add times and calendar

raoufchebri:


stepashka:
can we describe in more detail the format and how much time the event would take?
we need to set the expectations for our users, so that they know how to plan their time

PR #370, opened 2022-12-07
PITR update

danieltprice:
- Use "point-in-time restore" consistently. After some consideration, I think this term better reflects what you can do with branching. 
- Stop using the PITR acronym. It's used heavily for "point in time recovery", which means something different.


m-abdelwahab (approved):
I share the same feeling about the term restore. Looks good to me 👍

PR #380, opened 2022-12-08
Move connect section up

danieltprice:
Move the connect section before the mange section


danieltprice:
@bolotskydev or @vannyle 
Could I get an approval to merge? Thank you!

PR #399, opened 2022-12-14
Add import from csv instructions

danieltprice:
Add import from csv instructions
https://websitemain62807-dpriceimportfromcsv.gatsbyjs.io/docs/import/import-from-csv/


content/docs/import/import-from-csv.md
```diff
@@ -0,0 +1,52 @@
...
+   <Admonition type="tip">
+   You can also create tables using the **SQL Editor** in the Neon Console. See [Query with Neon's SQL Editor](../../query-with-neon-sql-editor).
+   </Admonition>
+
+3. Prepare a `customer.csv` file with the following data:
+
+   ```text
+   First Name,Last Name,Email
+   Casey,Smith,casey.smith@example.com
+   Sally,Jones,sally.jones@example.com
```
m-abdelwahab, line 43: Not having the index at the start resulted in an error:
```bash
ERROR:  invalid input syntax for type integer: "Casey"
```

After adding the index, the command runs successfully and returns the number of records copied to the database
```bash
COPY 2
```
```suggestion
   1,Casey,Smith,casey.smith@example.com
   2,Sally,Jones,sally.jones@example.com
```

content/docs/import/import-from-csv.md
```diff
@@ -0,0 +1,52 @@
...
+   1,Casey,Smith,casey.smith@example.com
+   2,Sally,Jones,sally.jones@example.com
+   ```
+
+4. From your `psql` prompt, load the data from the `customer.csv` file using the `\copy` option.
+
+    ```bash
+    \copy customer FROM '/path/to/customer.csv' DELIMITER ',' CSV HEADER
+    ```
+
```
m-abdelwahab, line 51: Suggestion: after running commands, we can include what the output will look like. This gives the developer following the guide more confidence that they're doing things right. What do you think?

```suggestion
If the command runs successfully, it will return the number of copied records to the database:

```bash
COPY 2
```

```

m-abdelwahab (approved):
Looks good to me 🚀

PR #410, opened 2022-12-16
Dprice manage section general revisions

danieltprice:
- General revisions to update content to reflect the current state.
- General edits, corrections, etc.
https://websitemain62807-dpricemanagesectiongeneralrevi.gatsbyjs.io/docs/manage/overview/


content/docs/manage/users.md
```diff
@@ -4,12 +4,14 @@ enableTableOfContents: true
 isDraft: false
 ---
 
-In Neon, users are PostgreSQL users. Each Neon project is created with a default user that takes its name from your Neon account (the Google, GitHub, or partner account that you registered with). This user owns the default database (`neondb`) that is created in a project's root branch.
+In Neon, users are PostgreSQL users. Each Neon project is created with a default user that takes its name from your Neon account (the Google, GitHub, or partner account that you registered with). This user owns the default database (`neondb`) that is created in a project's root branch. Each project is also has a `web_access` PostgreSQL user, which is a system managed user used by the Neon [SQL Editor](../../get-started-with-neon/query-with-neon-sql-editor) and for [passwordless connections](../../connect/passwordless-connect). You cannot delete or modify this user.
```
m-abdelwahab, line 7:
```suggestion
In Neon, users are PostgreSQL users. Each Neon project is created with a default user that takes its name from your Neon account (the Google, GitHub, or partner account that you registered with). This user owns the default database (`neondb`) that is created in a project's root branch. Each project also has a `web_access` PostgreSQL user, which is a system-managed user used by the Neon [SQL Editor](../../get-started-with-neon/query-with-neon-sql-editor) and for [passwordless connections](../../connect/passwordless-connect). You cannot delete or modify this user.
```

content/docs/manage/users.md
```diff
@@ -4,12 +4,14 @@ enableTableOfContents: true
...
 
-Additional users can be created in a project's root branch or child branches. There is no limit to the number fo users you can create.
+Additional users can be created in a project's root branch or child branches. There is no limit to the number of users you can create.
+
+Users belong to branch. If you create a child branch, users from the parent branch are duplicated in the child branch. For example, if user `sally` exists in the parent branch, user `sally` is copied to the child branch when the child branch is created. The only time this does not occur is when you create a branch that only includes data up to a particular point in time. If the user was created in the parent branch after that point in time, that user is not duplicated the child branch.
```
m-abdelwahab, line 11:
```suggestion
Users belong to a branch. If you create a child branch, users from the parent branch are duplicated in the child branch. For example, if user `sally` exists in the parent branch, user `sally` is copied to the child branch when the child branch is created. The only time this does not occur is when you create a branch that only includes data up to a particular point in time. If the user was created in the parent branch after that point in time, that user is not duplicated the child branch.
```

m-abdelwahab (approved):

PR #422, opened 2022-12-29
Update SNI info for Elixir Postgrex client

anirudhsama:
Clarified that Postgrex supports SNI, and also added a link to the official documentation which shows how it can be done.


danieltprice (approved):
Thank you for the contribution, @anirudhsama.

PR #435, opened 2023-01-06
Add a secure connections topic

danieltprice:
Add a secure connections topic
https://websitemain62807-dpriceaddsecureconnectionconte.gatsbyjs.io/docs/security/secure-connections/


lassizci (dismissed):
Looks good to me :+1:

acervantes23 (dismissed):

arssher (approved):
Expanding on SNI seems a bit excessive here, as it is marginally related to security, but why not.

What I think should be added to this topic is recommendation to verify hostname on the client side, so client library should behave like libpq with `sslmode=verify-full`. AFAIU we use general certificate authorities (let's encrypt or whatever), so forging endpoint with different hostname is easy. CC @funbringer

danieltprice:
> What I think should be added to this topic is recommendation to verify hostname on the client side, so client library should behave like libpq with `sslmode=verify-full`.

Hi @arssher, would we need to provide a root certificate file for users to implement hostname verification on the client side? See: https://www.postgresql.org/docs/current/libpq-ssl.html#LIBQ-SSL-CERTIFICATES. I recall implementing something similar for MySQL. A link was provided for users to download a region-specific cert bundle: https://dev.mysql.com/doc/heatwave-aws/en/heatwave-aws-host-name-identity-verification.html

content/docs/security/secure-connections.md
```diff
@@ -0,0 +1,20 @@
...
+```bash
+$> psql postgres://sally:************@ep-white-thunder-123456.us-east-2.aws.neon.tech/neondb
+psql (15.0 (Ubuntu 15.0-1.pgdg22.04+1), server 15.1)
+SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, compression: off)
+Type "help" for help.
+
+neondb=>
+```
+
+In addition to enforcing SSL/TLS for all connections, Neon implements the SNI (Server Name Indication) extension of the TLS protocol to ensure that clients access the correct SSL certificate for the Neon endpoint compute instance they are connecting to. SNI is an extension of the TLS protocol. The `libpq` PostgreSQL client library includes SNI support, in version 14 and higher. Connection requests from clients that do not support SNI are not permitted. Please refer to the discussion in [Connect from old clients](../../https://neon.tech/docs/connect/connectivity-issues/) for possible workarounds for clients that do not support SNI.
```
lassizci, line 20: > Connection requests from clients that do not support SNI are not permitted.

This actually isn’t true. Clients that don’t support SNI are required to specify a connection parameter.

danieltprice: Thank you @lassizci. I removed that line.

arssher:
> need to provide a root certificate file for users to implement hostname verification on the client side?

This is evidently not needed, we use some general CAs to which OS trusts by default (let's encrypt or something). But I see that libpq doesn't use the system wide certs, so this requires some setup anyway (`cp /etc/ssl/certs/ca-certificates.crt ~/.postgresql/root.crt` on my machine) -- we should something on that that if we go there.

Perhaps I exaggerate the need for hostname verification. I guess it is better to address that in separate PR, with consultation of someone working on proxy, not really an expert here.

arssher (approved):

content/docs/security/secure-connections.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Secure connections
+---
+
+Neon requires that all connections use SSL/TLS encryption to ensure that data sent over the Internet cannot be viewed or manipulated by third parties. Connections that do not use SSL/TLS are rejected. 
+
+Neon enforces SSL/TLS connections by enabling  `ssl` in PostgreSQL and using only `hostssl` records in the `pg_hba.conf` client authentication configuration file. A `hostssl` record matches connection attempts made using TCP/IP, but only if the connection is made with SSL/TLS encryption. For related information, see [The pg_hba.conf File](https://www.postgresql.org/docs/current/auth-pg-hba-conf.html), in the _PostgreSQL documentation_.
```
arssher, line 7: Let's remove this paragraph, see
https://neondb.slack.com/archives/C036U0GRMRB/p1673325779724649?thread_ts=1673301585.286539&cid=C036U0GRMRB
we can add 'behaving like usual PostgreSQL with hostssl records in pg_hba.conf file only' to the last sentence.

arssher (changes requested):

content/docs/security/secure-connections.md
danieltprice: Thank you. Revised as suggested.

arssher (approved):

PR #452, opened 2023-01-16
Update connectivity-issues.md

danieltprice:
- Update SNI error message to say "Endpoint ID" instead of "Project ID", since we tell users to connect via the Endpoint ID now. The error message was changed by this PR: https://github.com/neondatabase/neon/pull/3349.
https://websitemain62807-dpriceupdatesnierrormessage.gatsbyjs.io/docs/connect/connectivity-issues/
- Update table of drivers that shows SNI support.
  - pg8000 is now supported
  - PostgresClientKit actually does not support SNI (we incorrectly documented that it does support SNI)

### DIFF
--- a/content/docs/connect/connectivity-issues.md
+++ b/content/docs/connect/connectivity-issues.md
@@ -8,7 +8,7 @@ redirectFrom:
 In most cases, copying a connection string from the Neon **Dashboard** and using it in your project should work as is. However, with older clients and some native PostgreSQL clients, you may receive the following error:
 
 ```txt
-ERROR: The project ID is not specified. Either upgrade the PostgreSQL client library (libpq) for SNI support or pass the project ID (the first part of the domain name) as a parameter: '&options=project%3D'. See [https://neon.tech/sni](https://neon.tech/sni) for more information.
+ERROR: The endpoint ID is not specified. Either upgrade the PostgreSQL client library (libpq) for SNI support or pass the endpoint ID (the first part of the domain name) as a parameter: '&options=project%3D'. See [https://neon.tech/sni](https://neon.tech/sni) for more information.
 ```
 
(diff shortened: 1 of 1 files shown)


content/docs/connect/connectivity-issues.md
```diff
@@ -97,9 +97,9 @@ Native client libraries:
 | postgres.js       | JavaScript  | no                                                       |
 | pgmoon            | Lua         |                                                          |
 | asyncpg           | Python      | yes                                                      |
-| pg8000            | Python      | [depends](https://github.com/neondatabase/neon/pull/2008) |
+| pg8000            | Python      | yes (as of pg8000 v1.29.3)                               |
```
bayandin, line 100: I've double-checked why the previous version we used didn't work, and it turned out the problem was in the `scramp` library (a dependency of `pg8000`).  The bug with channel binding was fixed in `scramp>=1.4.3`, so now even `pg8000==1.29.1` will work (if `scramp>=1.4.3`), the later version of `pg8000` has such `scramp` version restriction

content/docs/connect/connectivity-issues.md
```diff
@@ -97,7 +97,7 @@ Native client libraries:
 | postgres.js       | JavaScript  | no                                                       |
 | pgmoon            | Lua         |                                                          |
 | asyncpg           | Python      | yes                                                      |
-| pg8000            | Python      | yes (as of pg8000 v1.29.3)                               |
+| pg8000            | Python      | yes (requires [scramp >= v1.4.3](https://pypi.org/project/scramp/), which is included in [pg8000 v1.29.3](https://pypi.org/project/scramp/) and higher)  |
```
bayandin, line 100:
```suggestion
| pg8000            | Python      | yes (requires [scramp >= v1.4.3](https://pypi.org/project/scramp/), which is included in [pg8000 v1.29.3](https://pypi.org/project/pg8000/) and higher)  |
```

bayandin (approved):
Looks good, thanks @danieltprice!

PR #453, opened 2023-01-17
Draft release notes for storage release

lubennikovaav:
https://websitemain62807-releasenotes1701.gatsbyjs.io/docs/release-notes/


content/release-notes/2023-01-17.md
```diff
@@ -0,0 +1,30 @@
...
+Update client libraries and runtimes for Postgres libraries we test.
+- `pg8000` works with Neon now tada
+- `PostgresClientKit` still doesn't support SNI
+
+https://github.com/neondatabase/neon/pull/3365/commits/4c6b507472a985bf18e78adf0c3a5f4f922aa7ac
+
+
+- Compute: Add prefetch stat to EXPLAIN output.
+
+https://github.com/neondatabase/neon/pull/3365/commits/1983c4d4ade037a0ce66333fa54cc7d0707d936c
```
vadim2404, line 30: @Daniel-ef 

We can create an additional article devoted to prefetch and what it means in our reality.

vadim2404 (dismissed):

danieltprice (approved):
lgtm!

content/release-notes/2023-01-17.md
vadim2404:
```sql
EXPLAIN (PREFETCH) SELECT * FROM ANY_TABLE
```

As an example

```
neondb=> explain (prefetch) select 1;
                     QUERY PLAN
----------------------------------------------------
 Result  (cost=0.00..0.01 rows=1 width=4)
   Prefetch: hits=0 misses=0 expired=0 duplicates=0
(2 rows)
```

But it's better to create a table, generate data into it, and run the query that scans the table sequentially

vadim2404: and a keynote is that the `PREFETCH` option in `EXPLAIN` is only available in Neon. It's not a standard PostgreSQL behavior.

PR #457, opened 2023-01-19
File_fdw extension support notes

vadim2404:


danieltprice:
Hi @vadim2404 Are the files wiped out because they are held in a session context?

vadim2404:
> Hi @vadim2404 Are the files wiped out because they are held in a session context?

The file can stay on the server only when compute is up-and-running. When we scale it to zero, all these files will be wiped out

danieltprice (approved):
Made a few revisions.

PR #474, opened 2023-01-26
Update 2023-01-10.md

danieltprice:
Revisit the latest storage release note to make sure users understand the feature and changes. There are some questions I need help with. Thank you!

- In terms users can understand, what are "layer files"?
- What do we mean by "local state loss"? Is this a Neon Pageserver failure?
- In general, how and why is cold storage important? 
- Do we cold-store data that has not been used after a period of time? What period of time is that?

### DIFF
--- a/content/release-notes/2023-01-10.md
+++ b/content/release-notes/2023-01-10.md
@@ -5,3 +5,9 @@ label: 'Storage'
 ### What's new
 
 - Pageserver: Added support for on-demand download of layer files, which enables fast redistribution of tenant data across Pageservers and recovery from local state loss. This feature augments Neon's cold data storage capability by allowing the data to be restored when it is needed.
+
+Question about the entry above:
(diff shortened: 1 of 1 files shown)


shanyp:
> * n terms users can understand, what are "layer files"?
Users probably don't care, but maintainers and people looking at the code base will
> * What do we mean by "local state loss"? Is this a Neon Pageserver failure?
yes if the local disk of the pageserver is wiped/malfunctioning or for any reason it needs to be recreated
> * In general, how and why is cold storage important?
cold storage is the db data packaged in Neon's layer file structure which resides in S3. It is being transferred to the pageserver whenever it's needed 
> * Do we cold-store data that has not been used after a period of time? What period of time is that?
currently, there is no expiration for it ,but there will be in the future

shanyp (approved):

PR #490, opened 2023-01-30
Fix links broken by the Vercel migration

danieltprice:
Fix links broken by the Vercel migration. Relative links are not supported.


next/content/docs/manage/branches.md
```diff
@@ -9,11 +9,11 @@ redirectFrom:
...
 
 A child branch is a copy-on-write clone the data in the parent branch. You can modify the data in a branch without affecting the data in the parent branch.
-For more information about branches and how you can use them in your development workflows, see [Branching](../../conceptual-guides/branching).
+For more information about branches and how you can use them in your development workflows, see [Branching](/docs/introduction/branching).
 
 You can create and manage branches using the Neon Console or [Neon API](https://neon.tech/api-reference/v2/). This topic covers both methods.
 
-Before you can create a branch, you must have a Neon project. If you do not have a Neon project, see [Create a project](../projects/#create-a-project).
+Before you can create a branch, you must have a Neon project. If you do not have a Neon project, see [Create a project](/docs/projects/#create-a-project).
```
sharnoff, line 16: This link gives a 404. Probably intended to be `/docs/manage/projects#create-a-project`?

sharnoff (approved):
Changes all look good - mentioned the one that gives a 404.

Non-blocker: I noticed that a lot of the links have trailing slashes before a hash -- like `/path/to/page/#header`. These just redirect to the page without a trailing slash. Unless there's some other reason I'm not aware of, maybe these should be changed?

next/content/docs/manage/branches.md
danieltprice: > These just redirect to the page without a trailing slash. Unless there's some other reason I'm not aware of, maybe these should be changed?

Fixed these in a new PR. Thanks for point that out.

PR #492, opened 2023-01-30
Update links 

danieltprice:
Update links after Vercel migration


danieltprice:
Conflicts with existing fixes.

PR #526, opened 2023-02-16
Communicate breaking changes better

nikitakalyanov:
I think it's important to communicate breaking changes, that's why I update the release notes. Some of this is based on user feedback. I do understand that updating the old release notes is of limited value but I still think it is good for general awareness/making a habit/historical perspective


danieltprice (approved):
lgtm!

danieltprice (approved):
approved

PR #548, opened 2023-03-01
update non-pooled connection string example

m-abdelwahab:
Update the connection string example for non-pooled connections


danieltprice (approved):

danieltprice:
@m-abdelwahab 
Approved this then I realized that I just fixed it in another PR that I pushed this morning.

PR #564, opened 2023-03-06
Update vercel.md

danieltprice:
Add note about support for connecting more than one project.
https://neon-next-git-dprice-vercel-doc-update-neondatabase.vercel.app/docs/guides/vercel#add-the-neon-vercel-integration

### DIFF
--- a/content/docs/guides/vercel.md
+++ b/content/docs/guides/vercel.md
@@ -31,7 +31,7 @@ When you push changes to the GitHub repository associated with your Vercel proje
 This topic describes how to add the Neon Vercel integration to your Vercel project.
 
 <Admonition type="important">
-The Neon Vercel integration connects one Vercel project to one Neon project. The integration does not currently support sharing a PostgreSQL database (a Neon project) between two or more Vercel projects, nor does it support connecting a single Vercel project to two PostgreSQL databases (two Neon projects).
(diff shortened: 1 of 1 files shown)


content/docs/guides/vercel.md
```diff
@@ -31,7 +31,7 @@ When you push changes to the GitHub repository associated with your Vercel proje
 This topic describes how to add the Neon Vercel integration to your Vercel project.
 
 <Admonition type="important">
-The Neon Vercel integration connects one Vercel project to one Neon project. The integration does not currently support sharing a PostgreSQL database (a Neon project) between two or more Vercel projects,  nor does it support connecting two PostgreSQL databases (Neon projects) to a single Vercel project.
+The Neon Vercel integration connects one Vercel project to one Neon project. The integration does not currently support sharing a PostgreSQL database (a Neon project) between two or more Vercel projects, nor does it support connecting a single Vercel project to two PostgreSQL databases (two Neon projects).
```
duskpoet, line 34: not sure if we should write, that we don't support connecting to two databases, nothing prevents users from adding any number of databases via custom env vars. We only don't support multiple Neon projects. So I would avoid writing about PostgreSQL databases and stick to Neon projects

stepashka: agree with Ed, let's stick to Neon Projects in this

danieltprice: Thank you. Updated as suiggested.

duskpoet (approved):

PR #580, opened 2023-03-15
Remove mention of Technical Preview

danieltprice:
Drop all mention of "Technical Preview" throughout the docs


danieltprice:
we're launching paid tiers but staying in technical preview

PR #591, opened 2023-03-16
Add category filter to release notes

yasell:
This PR adds a category filter to the release notes – https://neon-next-git-release-notes-categories-neondatabase.vercel.app/docs/release-notes


bolotskydev:
@yasell something is fishy here, link of type https://neon-next-git-release-notes-categories-neondatabase.vercel.app/docs/release-notes/storage do not work for me

danieltprice:
Hi @yasell please note that we just changed the "Serverless" tag to "Drivers" and added a "Plans" tag. More additions / changes are possible in the future. We can handle those as they come, though.

bolotskydev:
@danieltprice please, check out the preview and share your thoughts. One note regarding the grouping: it doesn't seem wise to write multiple releases in a single file, because it requires additional logic and makes it harder to work with the source files. In order to enable grouping, we are suggesting to change name pattern for release notes from `date` to `date-category`. This way we keep each release note whatever the category in separate files, but it makes it possible to proceed with ease and assign to the same date multiple releases. In the UI it is going to look like this:
<details>
<summary>Example</summary>
image
</details>
And in this case we'll move the link to a particular note from date at the left side to the bottom of each release notes entry
wdyt?

bolotskydev (approved):

PR #671, opened 2023-04-09
fix: compute-lifecycle spelling

mehdidotcafe:


danieltprice (approved):
Thank you for the fix!

PR #676, opened 2023-04-12
chore: add info about `pg_stat_statements` extension

vadim2404:


content/docs/extensions/pg-extensions.md
```diff
@@ -9,65 +9,66 @@ Neon supports the PostgreSQL extensions shown in the following table. For extens
 
 <a id="default-extensions/"></a>
 
-| Extension                | Extension Version (PostgreSQL 14)                               | Extension Version (PostgreSQL 15)                                |Note                                                                                                                |
```
vadim2404, line 12: @danieltprice , I added here only one line, but it seems that my editor decided to format the whole file :( Feel free to update it

danieltprice:
Thanks, Vadim. This item is included in the console release notes that were just published, so I think this PR can be closed. A couple of notes:
- Announcement of the new extension is under the 4-11 console release notes, as a "Control Plane" item. This breaks the pattern, but ..
- I need to start placing "Compute" release notes into their own category in the release notes (Neon separates compute and storage, so it makes no sense to combine compute and storage in the release notes). I will add a "Compute" category to the release notes and place all "Compute items under it, including the release note for the new extension. Hope to get this done very shortly.
- Maybe we need an "Extensions" category in the release notes to give them more visibility .. lmk what you thiink.

vadim2404:
oh didn't notice about control plane release note. It makes sense, then

PR #677, opened 2023-04-12
Highlight Prisma connection timeout instructions

danieltprice:
Update Prisma docs to emphasize adjusting your connection string to avoid timeout issues.
- Add link to connection timeout issues from the intro.
- Add a warning that mentions adjusting the connection string to avoid connection timeout issues.

Preview:
https://neon-next-git-dprice-prisma-connections-update-neondatabase.vercel.app/docs/guides/prisma

### DIFF
--- a/content/docs/guides/prisma.md
+++ b/content/docs/guides/prisma.md
@@ -36,9 +36,7 @@ To establish a basic connection from Prisma to Neon, perform the following steps
 
 3. Add a `DATABASE_URL` variable to your `.env` file and set it to the Neon connection string that you copied in the previous step. Your setting will appear similar to the following:
 
-   ```text
-   DATABASE_URL="postgres://daniel:<password>@ep-restless-rice-862380.us-east-2.aws.neon.tech/neondb"
-   ```
+   `DATABASE_URL="postgres://daniel:<password>@ep-restless-rice-862380.us-east-2.aws.neon.tech/neondb"`
 
 <Admonition type="important">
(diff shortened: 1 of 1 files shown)


m-abdelwahab (dismissed):

jdrumgoole:
This string `DATABASE_URL=postgres://daniel:<password>@ep-restless-rice-862380-pooler.us-east-2.aws.neon.tech/neondb?pgbouncer=true` is truncated in the dialog box so the pgbouncer part is hidden. Can you change the layout so the relevant parts of the URL are in view. Also can you bold the pieces that represent the changes e.g. "-pooler" and "?pgbouncer=true"?

danieltprice:
> This string `DATABASE_URL=postgres://daniel:<password>@ep-restless-rice-862380-pooler.us-east-2.aws.neon.tech/neondb?pgbouncer=true` is truncated in the dialog box so the pgbouncer part is hidden. Can you change the layout so the relevant parts of the URL are in view. Also can you bold the pieces that represent the changes e.g. "-pooler" and "?pgbouncer=true"?

Hi Joe,
I've tried a number of ways to implement what you suggested, but our Markup and parser won't support showing all of the connection string in code sample format with parts in bold. The best I could do is format the connection string as in in-line code snippet to have the full connection string appear on the page.  

I'd like to be able to present the code block/snippets as you suggested, so I opened a task for Pixel Point to implement something: https://github.com/orgs/neondatabase/projects/10?pane=issue&itemId=25476676

In the meantime, take a look at how the connection string examples look as in-line code snippets instead of code blocks and let me know if you think it's an improvement. You can see all of the connection string, at least.

I suggest we do what we can now until Pixel Point is able to implement a change, which might be a couple of weeks out.

m-abdelwahab (approved):

PR #711, opened 2023-05-01
Improve and shorten Go connect example code

tmaxmax:
Hopefully this is not too pedantic of a change, but here it is. I updated the sample Go code for connecting to and querying the database to reflect idiomatic code:
- it does not use `log.Fatal` in conjunction with `defer`: `log.Fatal` calls `os.Exit`, which abruptly ends the program without calling any deferred functions
- it uses `QueryRow` instead of `Query`, which should in general be used for querying a single value; the resulting code is much shorter

By having the example shorter, readers can focus better and mainly on the connection part.

BTW, great product and great service! Thank you for creating this.

### DIFF
--- a/content/docs/guides/go.md
+++ b/content/docs/guides/go.md
@@ -38,7 +38,6 @@ package main
 import (
     "database/sql"
     "fmt"
-    "log"
 
     _ "github.com/lib/pq"
 )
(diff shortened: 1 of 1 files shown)


danieltprice (approved):
Thank you for the contribution, @tmaxmax!

PR #729, opened 2023-05-10
Dprice add sequelize instructions

danieltprice:
Add instructions for connecting from Sequelize. Users get stuck on a require ssl error. This guide addresses that. 
https://neon-next-git-dprice-add-sequelize-instructions-neondatabase.vercel.app/docs/guides/sequelize


danieltprice:
we've covered it in another guide

PR #742, opened 2023-05-12
fix: release notes paths

vannyle:
This pull request brings the fixes:
- replace relative path with absolute
- SEO title


danieltprice:
Hi @vannyle The absolute links do not go to the anchored subsection. Only the relative links seem to work.
Try one of the glossary terms for example:
/docs/reference/glossary#pageserver

PR #752, opened 2023-05-17
feat: added prices json and script to sync it with orb

duskpoet:


scripts/get_prices.js
```diff
@@ -0,0 +1,59 @@
+#! /bin/env node
+
+const http = require('https');
```
seymourisdead, line 3: I'd rather had it in the cloud repo, because it is not necessary to expose it to the public.

Also we can add the script to the CI and upload the result somewhere and we can use the result both for the Console UI and for the landing and for other places too.

personally i would say that my favorite way to do this was the prices API endpoint

duskpoet: It should be public, so it could be used from the landing website

duskpoet: It is a public information I see no point in hiding it

seymourisdead: the prices are public, the process how we retrieve them - not necessarily

duskpoet: oh, I didn't notice this is a public repo ...

PR #778, opened 2023-05-26
Release notes refresh

danieltprice:
Move developer-oriented content out of the Release notes. The Release Notes will become more user-oriented and focused on benefits and usage of the new features or changes. Preview:
https://neon-next-git-dprice-release-notes-refresh-neondatabase.vercel.app/docs/release-notes

Developer-oriented storage and compute content will be moved to a CHANGELOG.md file in the neon repository. This PR has a manual example:
https://github.com/neondatabase/neon/pull/4365

Vadim is looking into an automated solution.


shanyp:
Suggestion: drop "Storage Compute" in favor of "Core" or "Engine". As "Storage Compute" is just how we structure teams internally

danieltprice:
> Suggestion: drop "Storage Compute" in favor of "Core" or "Engine". As "Storage Compute" is just how we structure teams internally

Good point @shanyp  I agree. We should consider dropping any categorizations that look inward. We need to change to a user's perspective. This change will require a redesign involving Pixel Point. Let's aim to drop or revise the categorizations in the next release notes redesign.

ololobus (commented):
I'm not a fan of editing release notes retrospectively if it's not some typos fixing or rephrasing

danieltprice:
> I'm not a fan of editing release notes retrospectively if it's not some typos fixing or rephrasing

This is a fair point. I'd like to move us to a new format for the release notes, which would require restructuring at the very least. I'll have to think about how we could make this transition without altering the existing record.

danieltprice:
We'll leave the existing release notes as-is. We'll make a break with the new design.

PR #798, opened 2023-06-01
Add cold start topic

danieltprice:
Add cold start topic. Preview:
https://neon-next-git-dprice-manage-cold-starts-neondatabase.vercel.app/docs/connect/cold-starts


content/docs/connect/cold-starts.md
```diff
@@ -0,0 +1,198 @@
...
+
+Here are examples of how to increase connection timeout settings in a few common programming languages and frameworks:
+
+<CodeTabs labels={["Node.js", "Python", "Java", "Prisma" ]}>
+
+```js
+const { Pool } = require('pg')
+
+const pool = new Pool({
+connectionString: process.env.DATABASE_URL,
```
skyzh, line 59: The indent does not look correct?

danieltprice: Thank you. Fixed

acervantes23:
@danieltprice I think whole paragraph should be rephrased in a more positive tone

`With serverless architectures and auto-scaling systems such as Neon, a common challenge that arises is "cold starts". For Neon, a cold start refers to the situation where a compute instance is started in response to a request after being in an idle state. Cold starts can result in longer response times and connection failures or timeouts due to the few seconds required to restart an idle compute.`

The issue I see with it is that it points to cold starts as a problem and only focuses on the problem. I would put the emphasis more on `optimizing performance`...

danieltprice:
> The issue I see with it is that it points to cold starts as a problem and only focuses on the problem. I would put the emphasis more on `optimizing performance`...

Thank you. Makes sense. I will revise.

acervantes23 (dismissed):

danieltprice:
Closing. I decided to take a different approach to this content considering that users will look for it as they experience connection latencies or timeouts. Please see https://github.com/neondatabase/website/pull/812 for the revised version.

PR #822, opened 2023-06-12
chore: performance improvements for docs introduction

vannyle:
This pull request brings the performance improvements for docs:
- set explicit image size on the docs introduction page
- updated breadcrumb rendering condition 
- replaced chevrons in the sidebar as inline SVG to mask-image to reduce the DOM size from 1581 elements to 1380 elements, it’s not an error, but still a warning. (The DOM size should be fewer than 800 elements)
- there is an `<input type=“text” />` was added as a workaround to prevent scroll in Safari when closing the Docsearch’s modal, added `aria-label=“hidden”` to fix the error: “Form elements do not have associated labels”.
- added a prop `ariaLabel` to the sidebar item, as there is a link “Go” in the sidebar, the Lighthouse considers it as a generic link text (click here, click this, go, here, this, start, right here, more, learn more)

**Additional changes**
- set `layout` and `footer` components as server components


next.config.js
```diff
@@ -14,6 +14,8 @@ module.exports = {
         hostname: '**',
       },
     ],
+    deviceSizes: [640, 750, 760, 828, 928, 1080, 1200, 1432, 1592, 1920, 2048, 3840],
+    imageSizes: [424],
```
lnikell, line 18: Can you elaborate more regarding the purpose of increasing the number of device sizes from a default one, as well as decreasing the number of default image sizes?

src/components/pages/doc/sidebar/item/item.jsx
```diff
@@ -40,19 +46,28 @@ const Item = ({ title, slug = null, isStandalone = null, items = null, currentSl
...
+              ? 'bg-black-new dark:bg-white'
+              : 'bg-gray-new-40 group-hover:bg-black-new dark:bg-gray-new-90 dark:group-hover:bg-white',
             items?.length ? 'block' : 'hidden',
             isOpen ? 'rotate-90' : 'rotate-0'
           )}
+          style={{
+            WebkitMaskImage: "url('/images/chevron-right.svg')",
+            WebkitMaskSize: '6px 10px',
+            WebkitMaskRepeat: 'no-repeat',
+            WebkitMaskPosition: 'center',
```
lnikell, line 69: Very interesting solution, can you tell more about how does it work?

vannyle: As we need to color the SVGs, one of the easy ways to do it is using a CSS mask.

next.config.js
vannyle: I added the sizes for testing, already removed them

src/components/pages/doc/sidebar/item/item.jsx
vannyle: https://codepen.io/noahblon/post/coloring-svgs-in-css-background-images#css-masks-1

PR #830, opened 2023-06-14
Update security.txt

danieltprice:
Update link in security.txt. The location of the page changed.


vadim2404 (approved):

public/security.txt
```diff
@@ -1,4 +1,4 @@
 Contact: mailto:security@neon.tech
 Expires: 2022-12-31T23:59:00.000Z
```
vadim2404, line 2: I guess we need to update this date

vadim2404 (changes requested):

public/security.txt
danieltprice: Thank you. I missed that. Updated. 

For my future reference:
"The Expires field in a security.txt file specifies the date and time when the security policy described in the security.txt file expires, and it is set by you, the creator of the file. The purpose of this field is to avoid having outdated security policies in place. If the date has already passed, the security.txt should be considered expired.

To update it, you simply need to change the Expires field to a future date and time, preferably in the ISO 8601 format (YYYY-MM-DDThh:mm:ss.sTZD), e.g., 2023-12-31T23:59:00.000Z. This will give anyone referring to your security.txt file the certainty that your contact information, policy, and other provided details are current and valid.

After setting a new expiration date, make sure to review all the other information in the file to ensure everything is still correct and up to date. 

Lastly, remember to regularly review and update your security.txt file, at least once per year or more frequently if necessary.

vadim2404 (approved):

PR #847, opened 2023-06-23
refactor: minor refactoring for docs

vannyle:
This pull request brings minor refactoring:
- removed importing react
- renamed heading `h3` in footer into `span` (avoid the error when heading hierarchy out of order
- set `layout` and `footer` components as server components
- removed IntersectionObserver for StatusBadge in footer


sashasushko (commented):
LGTM ❤️ Please resolve the conflicts and make sure that every client component in home page market as `'use client'`

PR #861, opened 2023-07-05
Dprice pro tier limits update

danieltprice:


danieltprice:
Addressed in a different PR.

PR #868, opened 2023-07-07
Update aws-lambda.md

ffxsam:
Using this approach (declaring client outside the handler func) ensures that the client connection is persisted across warm starts. This prevents too many new connections from being instantiated to the NEON DB.

Also, `client.connect()` returns a Promise, and that was overlooked in the previous example.

### DIFF
--- a/content/docs/guides/aws-lambda.md
+++ b/content/docs/guides/aws-lambda.md
@@ -150,7 +150,7 @@ Create the Lambda function using the [Serverless Framework](https://www.serverle
 
     module.exports.getAllUsers = async () => {
      if (!client) {
-       var client = new Client(process.env.DATABASE_URL);
+       client = new Client(process.env.DATABASE_URL);
        await client.connect();
      }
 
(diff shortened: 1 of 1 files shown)


content/docs/guides/aws-lambda.md
```diff
@@ -146,11 +146,16 @@ Create the Lambda function using the [Serverless Framework](https://www.serverle
...
     const { Client } = require('pg');
+    let client;
 
     module.exports.getAllUsers = async () => {
-     var client = new Client(process.env.DATABASE_URL);
-     client.connect();
-     var { rows } = await client.query('SELECT * from users');
+     if (!client) {
+       client = new Client(process.env.DATABASE_URL);
+       await client.connect();
```
rishi-raj-jain, line 154: even if the client is present and being re-used, the connection still needs to be ensured using .connect(), correct?

dhanushreddy291:
#2841

danieltprice:
Thank you for the PR. We believe the issue has been addressed by #2841. Please let us know if anything remains to be addressed.

PR #877, opened 2023-07-12
feat: og image generation for docs

vannyle:
This pull request brings the og image generation for docs

**Screenshots**
image

**Step to steps**
- go to https://socialsharepreview.com/
- check any doc page https://neon-next-git-og-image-neondatabase.vercel.app/docs/introduction


src/app/docs/[...slug]/page.jsx
```diff
@@ -40,14 +40,14 @@ export async function generateMetadata({ params }) {
...
-    data: { title, ogImage },
+    data: { title },
     excerpt,
   } = post;
 
   return getMetadata({
     title: `${title} - Neon Docs`,
     description: isReleaseNotes ? 'The latest product updates from Neon' : excerpt,
-    imagePath: ogImage,
+    imagePath: `https://neon-next-git-og-image-neondatabase.vercel.app/api/og-image?title=${title}`,
```
Comment (lnikell) [inline src/app/docs/[...slug]/page.jsx:50]: Let's use vercel system env variables to have ability to test og images in preview and non preview environments.

It feels like we can use `VERCEL_ENV` to check whether it's preview or not preview environment. If it's preview we can build a url for it using `VERCEL_BRANCH_URL` , if not we can use our variable with primary domain.
https://vercel.com/docs/concepts/projects/environment-variables/system-environment-variables

src/pages/api/og-image.jsx
```diff
@@ -0,0 +1,139 @@
+import { ImageResponse } from 'next/server';
+
+export const runtime = 'edge';
```
lnikell, line 3: Let's try to avoid using old pages directory for this. Here is a example of how can you make it work with app dir.

https://github.com/leerob/leerob.io/blob/main/app/og/route.tsx

Also since og images are very sensitive to url changes, let's don't make that path(current /api/og-image) global if it's at this moment used only in docs. So I'd suggest to narrow it down.

src/app/docs/release-notes/[...slug]/page.jsx
```diff
@@ -30,25 +30,29 @@ export async function generateMetadata({ params }) {
   const { slug } = params;
 
   let label = '';
+  let releaseDate = '';
   let description = '';
   const currentSlug = slug.join('/');
   const isReleaseNotePage = RELEASE_NOTES_SLUG_REGEX.test(currentSlug);
```
Comment (lnikell) [inline src/app/docs/release-notes/[...slug]/page.jsx:36]: Let's not change anything related to release notes for the time being, since we will soon be working on a redesigned version for it. I think we will use a custom release notes og image instead of auto generated at the end.

lnikell (changes requested):

src/app/docs/release-notes/[...slug]/page.jsx
Comment (vannyle) [reply src/app/docs/release-notes/[...slug]/page.jsx:36]: Got it, thanks for letting me know

PR #879, opened 2023-07-12
Update Neon CLI connection content

danieltprice:
Add headings for the connection methods:
https://neon-next-git-dprice-update-cli-connect-content-neondatabase.vercel.app/docs/reference/neon-cli#connect


content/docs/reference/neon-cli.md
```diff
@@ -44,15 +44,21 @@ To upgrade to the latest version of the Neon CLI, run the `npm i -g neonctl` com
...
 neonctl auth
 ```
 
-The `auth` command launches a browser window where you can authorize the Neon CLI to access your Neon account. Running a Neon CLI command without authenticating with [neonctl auth](/docs/reference/cli-auth) automatically launches the browser authentication process.
+The [neonctl auth](/docs/reference/cli-auth) command launches a browser window where you can authorize the Neon CLI to access your Neon account. If you have not authenticated previously, running a Neon CLI command automatically launches the web authentication process unless you have specified an API key in your command.
+
+### Connect with an API key
 
-Alternatively, you can authenticate a connection with a Neon API key using the `--api-key` option when running a Neon CLI command. For example, an API key is used with the following `neonctl projects list` command:
+To connect with a Neon API key, specify the `--api-key` option when running a Neon CLI command. For example, the following `neonctl projects list` command connects to Neon using the `--api-key` option:
```
duskpoet, line 61: should we add info that cli looks at the NEON_API_KEY env as a first attempt? (sorry, it kind of slipped through)

duskpoet (approved):

content/docs/reference/neon-cli.md
danieltprice: Thanks, Eduard. I mentioned that the --api-key option uses the NEON_API_KEY value if set. Let me know if that's not quite accurate.

content/docs/reference/neon-cli.md
```diff
@@ -44,21 +44,29 @@ To upgrade to the latest version of the Neon CLI, run the `npm i -g neonctl` com
...
 
 ```bash
 neonctl projects list --api-key <neon_api_key>
 ```
 
-For information about obtaining an Neon API key, see [Authentication](https://api-docs.neon.tech/reference/authentication), in the _Neon API Reference_.
+For information about obtaining an Neon API key, see [Create an API key](https://neon.tech/docs/manage/api-keys#create-an-api-key).
+
+To avoid specifying an API key with each CLI command, you can export your API key to the `NEON_API_KEY` environment variable. The `--api-key` option uses the `NEON_API_KEY` value if set.
```
duskpoet, line 69: I think this could give the false impression that env variable overrides the option, it's another way around. The priority would be the following:

1. if `--api-key` is specified, that value is used, otherwise
2. if NEON_API_KEY env is set, that value is used, otherwise
3. if `<config-dir>/credentials.json` file is present, then store credentials are used. This file is created as a result of "auth" command. Otherwise
4. if environment doesn't look like CI, then auth flow is executed. Otherwise throw error

danieltprice: Thank you. Revised.

duskpoet (approved):

PR #907, opened 2023-07-23
Remove role and database access management example.

danieltprice:
Remove role and database access management example. We seem to have removed the ability to grant membership in another role (GRANT dev_users TO dev_user1;), which requires the ADMIN OPTION privilege.

```
neondb=> GRANT dev_users TO dev_user2;
ERROR:  permission denied to grant role "dev_users"
DETAIL:  Only roles with the ADMIN option on role "dev_users" may grant this role.
```

Updated neon_superuser section to indicate that roles in Neon cannot grant or revoke membership in other roles.

Note: 07/24. This issue is under investigation. It may not be an intended restriction.


danieltprice:
The issue will be fixed. This PR is no longer required.

PR #951, opened 2023-08-12
Update cli-branches.md

danieltprice:
Removes connection URI output from branches create examples


duskpoet (changes requested):
why? it's still there

danieltprice:
I think I had an older version of neonctl. I thought I updated. Maybe I needed to reopen my perminal. Not sure why I was not seeing the connect uri.

PR #974, opened 2023-08-25
Note ipv4 support

danieltprice:
Neon currently supports ipv4. Mention this in the docs.
https://neon-next-git-dprice-ipv4-support-neondatabase.vercel.app/docs/connect/connect-from-any-app#network-protocol-support


danieltprice:
Maybe we should mention here that we also offer a serverless driver that supports both WebSocket and HTTP connections?

atumark:
> Maybe we should mention here that we also offer a serverless driver that supports both WebSocket and HTTP connections?

Makes sense. Even though we have them mentioned under [serverless driver itself](https://neon.tech/docs/serverless/serverless-driver) I'd say it's worth referencing here as well.

atumark (dismissed):
lgtm!

PR #1014, opened 2023-09-12
Update the docs link to the Neon status page

danieltprice:
Update the docs link to the Neon status page


danieltprice:
We're keeping the same domain name.

PR #1026, opened 2023-09-20
Remove content directory from prettierignore

danieltprice:
Hi Kate, any reason not to remove the content directory from prettierignore? If not, can you approve this PR?


danieltprice:
@k8isdead Hi Kate, is the final word on this that we cannot remove the content directory from prettierignore or should I open a task for further investigation?

PR #1037, opened 2023-09-26
Last updated date for docs

vannyle:
This pull request brings the last updated date for docs.

### DIFF
--- a/content/docs/ai/ai-concepts.md
+++ b/content/docs/ai/ai-concepts.md
@@ -2,6 +2,7 @@
 title: AI Concepts
 subtitle: Learn about embeddings and how they are used to build AI and LLM applications
 enableTableOfContents: true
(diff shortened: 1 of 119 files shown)


k8isdead (approved):
LGTM

danieltprice:
Thank you for this.
Small change request: Could we right align this and make the font slight smaller than the regular text?

k8isdead:
@danieltprice we'll decrease the font size a bit, but I suggest we keep it left aligned since it's the most common option in terms of design and logic

danieltprice:
> @danieltprice we'll decrease the font size a bit, but I suggest we keep it left aligned since it's the most common option in terms of design and logic

ok. Let's go with that.

danieltprice (approved):

PR #1041, opened 2023-09-28
docs/update-connection-examples-images

PaulieScanlon:
- add x4 new dark mode images for sql editor and connection details
connection_details


danieltprice:
Hey Paul, Thanks for this. A couple of steps I'd like to consider and implement before merging:
I really like the style with the dark mode and green lines. I think we should standardize on that for all screen caps. It looks better! 
- The connection_details.png file is used in several topics in the docs, maybe in about 10 other places. I should review those to make sure we don't have a mix of light and dark mode screen captures in a single topic.
- The connection string in that screen cap in referenced in the surrounding text in some cases, so the docs will have to be updated.
If you don't mind, I'll commandeer this PR to make the necessary changes.

PaulieScanlon:
Yeap, no worries @danieltprice. Yes there's a few other places where light mode is used. Would you prefer I prepare those images in this PR, or make another one once this is merged?

danieltprice:
Since we've changed the connection strings, we'll retake these captures under a new PR.
Thank you for your work on this, Paul. Much appreciated.

PR #1052, opened 2023-10-03
docs/javascript-next-js-connection-snippets

PaulieScanlon:
- update nextjs, node and prisma snippets to include `sslmode`

I've added two different options for Next.js. One is for API Routes, the other is for `getServerProps`. Next and Node snippets should be pretty similar. The Prisma one is a little more clunky as the user is required to update the connection string, rather than add config options to the client.


danieltprice (approved):
LGTM!

PR #1053, opened 2023-10-03
docs/javascript-code-fence

PaulieScanlon:
- change all `js` to `javascript` for code fence lang

We were using both `js` and `javascript` for the code fence lang, both will work but just to be consistent I've changed them all to `javascript`... or I can change them to all to `js`?


danieltprice (approved):
`javascript` is good. We'll use that going forward. Thank you!

PR #1054, opened 2023-10-03
docs/user-in-connection-string

PaulieScanlon:
- change to the singular "user" in the connection string snippet.

Just a small one to change instances of `<users>` to `<user>`


danieltprice (approved):
Thank you!

PR #1057, opened 2023-10-04
docs/go-connection-snippets

PaulieScanlon:
## Update:

Squares boys have now been used. I've also removed the "explanation" for connection string parts. I'm thinking we point users to the "Connect from any application." then the breakdown will only ever live in one place in the docs. — What do you think @danieltprice / @bgrenon 

---
- add `sslmode=require` to go snippet

This is a good candidate for why we should perhaps rethink how we display connection strings. 

This, IMO is quite confusing.
```
connStr := "user=<user> password=<password> dbname=neondb host=<hostname>?sslmode=require"
```

It might make more sense if it were.

```
postgres://<user>:<password>@ep-snowy-unit-550577.us-east-2.aws.neon.tech/neondb?sslmode=require
```

### DIFF
--- a/content/docs/guides/go.md
+++ b/content/docs/guides/go.md
@@ -5,7 +5,7 @@ enableTableOfContents: true
 redirectFrom:
   - /docs/quickstart/go
   - /docs/integrations/go
-updatedOn: '2023-10-04T12:27:20.313Z'
+updatedOn: '2023-10-05T13:29:38.514Z'
 ---
 
 To connect to Neon from a Go application:
@@ -44,7 +44,7 @@ import (
 )
 
 func main() {
-    connStr := "user=<user> password=<password> dbname=neondb host=<hostname>?sslmode=require"
+    connStr := "postgres://[user]:[password]@[neon_hostname]/[dbname]?sslmode=require"
     db, err := sql.Open("postgres", connStr)
     if err != nil {
         panic(err)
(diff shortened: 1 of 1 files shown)


danieltprice:
I agree. Seems confusing. 

Just punching this into GPT, seems that we could do this:

func main() {
	psqlInfo := "postgres://<user>:<password>@<host>:5432/<dbname>?sslmode=require"

	db, err := sql.Open("postgres", psqlInfo)
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()

	err = db.Ping()
	if err != nil {
		log.Fatal(err)
	}
}


or concatenate the variables, like this:

const (
	host     = "<host>"
	port     = 5432
	user     = "<user>"
	password = "<password>"
	dbname   = "<dbname>"
)

func main() {
	psqlInfo := "postgres://" + user + ":" + password + "@" + host + "/" + dbname + "?sslmode=require"

	db, err := sql.Open("postgres", psqlInfo)
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()

	err = db.Ping()
	if err != nil {
		log.Fatal(err)
	}
}

The the first example seems like the simplest.

danieltprice (approved):

PaulieScanlon:
I think this might be another prime candidate for Square Boys vs Alex, as mentioned in #1066. Will carry on that conversation over there and circle back to this.

danieltprice (approved):

PR #1070, opened 2023-10-05
docs/java-connection-snippets

PaulieScanlon:
- update java connection string with `sslmode=require`

This is a weird one, should we perhaps use Alex instead of Square Boys @danieltprice?


danieltprice (approved):

danieltprice:
The brackets seem reasonable here.

PR #1076, opened 2023-10-05
docs/next-js-app-router-snippet

PaulieScanlon:
- add app router code snippets.

App Router is the "new" way to use Next.js so I've moved it to the top of the examples. It's followed by the "old" way using the Pages Router then finally, the API Routes example.

@danieltprice, I had a bit of trouble with Prettier and the code blocks, would you mind giving this a quick pull and seeing if auto-format works? I kept getting errors, but I couldn't figure out why.

### DIFF
--- a/content/docs/guides/nextjs.md
+++ b/content/docs/guides/nextjs.md
@@ -5,7 +5,7 @@ enableTableOfContents: true
 redirectFrom:
   - /docs/quickstart/vercel
   - /docs/integrations/vercel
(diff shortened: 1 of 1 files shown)


danieltprice (approved):
LGTM!
Just some line spacing issues, I think. All sorted.
Unfortunately, Prettier is not working with .md files at the moment. It's a known issue: https://github.com/orgs/neondatabase/projects/10/views/5?filterQuery=prettier&pane=issue&itemId=40105331
Thank you, Paul!

PR #1082, opened 2023-10-06
Update sqlalchemy.md

danieltprice:
Remove @ in variable, which will result in double @@

Hey Paul, can you validate my assumption here that we had an extra "@" in the hostname.


PaulieScanlon:
@danieltprice good catch. That would have indeed resulted in double `@`

PaulieScanlon (approved):

PR #1090, opened 2023-10-10
docs/next-js-static-props-snippet

PaulieScanlon:
- add code snippet for `getStaticProps`. This method is almost identical to `getServerSideProps` but I've added it for good measure. I'm not sure if these should be under their own heading though, what do you think @danieltprice?

### DIFF
--- a/content/docs/guides/nextjs.md
+++ b/content/docs/guides/nextjs.md
@@ -121,7 +121,7 @@ There are two methods for fetching data using server-side requests in Next.js th
 1. `getServerSideProps` fetches data at runtime so that content is always fresh.
 2. `getStaticProps` pre-renders pages at build time for data that is static or changes infrequently.
 
-### getServerSideProps
+#### getServerSideProps
 
 From `getServerSideProps` using the Pages Router, add the following code snippet to connect to your Neon database:
 
@@ -169,7 +169,7 @@ export default function Page({ data }) {}
 
 </CodeTabs>
 
-### getStaticProps
+#### getStaticProps
 
 From `getStaticProps` using the Pages Router, add the following code snippet to connect to your Neon database:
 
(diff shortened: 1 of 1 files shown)


content/docs/guides/nextjs.md
```diff
@@ -116,7 +116,7 @@ export default async function Page() {
 
 ### Pages Router
```
bgrenon, line 118: Depending on how often your data changes, you can use either of these methods for getting your data:
- `getServerSideProps` fetches data at runtime so that content is always fresh.
- `getStaticProps` pre-renders pages at build time for data that is static or changes infrequently.

#### `getServerSideProps`

content/docs/guides/nextjs.md
```diff
@@ -162,6 +162,52 @@ export default function Page({ data }) {}
 
 </CodeTabs>
```
bgrenon, line 164: #### `getStaticProps`

content/docs/guides/nextjs.md
danieltprice: Makes sense to me!

PaulieScanlon:
Thanks for the suggestions and feedback @danieltprice and @bgrenon.

I've made a change and given each `getS.. ` function its own heading however if I use semantic headings (these should be h4s or `####`), they don't appear in the list under "On this page".

I've set them as h3s or `###` for now. What do think?

danieltprice:
OK. I see what you mean. You want them to show up in the right TOC, it only shows h2 and h3 headings. 

image

Yeah, this is a tough one. A few options:
1. Define them as h4 and don't worry about them showing up in the right TOC. You should still be able to link to h4 headings, if that's a concern.
2. Change the title to "Pages Router methods"
3. Ask PP to enable another level of heading in the right TOC. Not sure how this will look. Would it be too busy.

I would probably do 1 _and_ 2 in this case, and not worry about h4 level items appearing in the right TOC.

PaulieScanlon:
Thanks @danieltprice. I've changed them to h4 and will drop PP a note to see if they can take a look at the TOC.

PR #1096, opened 2023-10-11
docs/astro-guide

PaulieScanlon:
- add astro guide

FYI. Astro uses Vite, rather than Node so the way you access environment variables is different. 

E.g This
```
// vite
import.meta.env.DATABASE_URL
```

Not this

```
// node
process.env.DATABASE_URL
```


danieltprice (approved):
Thank you, Paul! This is great. Glad to see us expanding our content here.

PR #1097, opened 2023-10-11
docs/add-serverless-to-next-js-guide

PaulieScanlon:
- add neon serverless driver snippets to all request methods

Hi @danieltprice, hi @bgrenon. This might end up being quite a large task because the changes I've made here might need to be rolled out to all other guides/languages. I've discussed this with the team and we think we should be providing examples for using the serverless driver wherever it can be used. To my knowledge, there's nowhere it can't be used and as such, it should sit alongside all `pg` and `postgres.js` code snippets.


danieltprice (approved):

danieltprice:
Makes perfect sense to add a serverless driver snippet to theses guides!

PR #1101, opened 2023-10-12
docs/remix-guide

PaulieScanlon:
-  add remix guide

### DIFF
--- a/content/docs/guides/remix.md
+++ b/content/docs/guides/remix.md
@@ -1,0 +1,124 @@
+---
+title: Connect a Remix application to Neon
+subtitle: Set up a Neon project in seconds and connect from a Remix application
+enableTableOfContents: true
+updatedOn: '2023-10-06T17:44:14.691Z'
+---
+
+Remix is a full stack web framework that lets you focus on the user interface and work back through web standards to deliver a fast, slick, and resilient user experience. People are gonna love using your stuff.
+
+
+To create a Neon project and access it from a Remix application:
+
+1. [Create a Neon project](#create-a-neon-project)
+2. [Create a Remix project and add dependencies](#create-a-remix-project-and-add-dependencies)
+3. [Configure a Postgres client](#configure-the-postgres-client)
+4. [Run the app](#run-the-app)
+
+## Create a Neon project
+
+If you do not have one already, create a Neon project. Save your connection details including your password. They are required when defining connection settings.
+
+1. Navigate to the [Projects](https://console.neon.tech/app/projects) page in the Neon Console.
+2. Click **New Project**.
+3. Specify your project settings and click **Create Project**.
+
+## Create a Remix project and add dependencies
+
+1. Create a Remix project if you do not have one. For instructions, see [Quick Start](https://remix.run/docs/en/main/start/quickstart), in the Remix documentation.
+
+2. Add project dependencies using one of the following commands:
+
+    <CodeTabs labels={["node-postgres", "postgres.js", "Neon serverless driver"]}>
+
+      ```shell
+      npm install pg
+      ```
+
+      ```shell
+      npm install postgres
+      ```
+
+      ```shell
+      npm install @neondatabase/serverless
+      ```
+
+    </CodeTabs>
+
+## Store your Neon credentials
+
+Add a `.env` file to your project directory and add your Neon connection string to it. You can find the connection string for your database in the **Connection Details** widget on the Neon **Dashboard**. For more information, see [Connect from any application](/docs/connect/connect-from-any-app).
+
+<CodeBlock shouldWrap>
+
+```shell
+DATABASE_URL=postgres://[user]:[password]@[neon_hostname]/[dbname]
+```
+
+</CodeBlock>
+
+## Configure the Postgres client
+
+Create a `db.server.ts` or `.js` file at the root of your `app` directory add the following code snippet to connect to your Neon database:
+
+<CodeTabs labels={["node-postgres", "postgres.js", "Neon serverless driver"]}>
+
+```javascript
+import pg from 'pg';
+
+const pool = new pg.Pool({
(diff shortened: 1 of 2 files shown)


content/docs/guides/remix.md
```diff
@@ -0,0 +1,189 @@
+---
+title: Connect a Remix application to Neon
+subtitle: Set up a Neon project in seconds and connect from a Remix application
+enableTableOfContents: true
+updatedOn: '2023-10-06T17:44:14.691Z'
+---
+
+Remix is a full stack web framework that lets you focus on the user interface and work back through web standards to deliver a fast, slick, and resilient user experience. People are gonna love using your stuff.
```
brookslybrand, line 8: Would you be open to making this a quote and linking to our website since this is the h1 copy?

PaulieScanlon: @danieltprice, that sounds like one for you.

content/docs/guides/remix.md
```diff
@@ -0,0 +1,189 @@
...
+
+</CodeBlock>
+
+## Configure the Postgres client
+
+TODO. Add explanation for why a `.db.server` file is required.
+
+### db.server
+
+Create a `db.server.ts` file at the root of your `/app` directory add the following code snippet to connect to your Neon database:
```
brookslybrand, line 66: Totally just a suggestion since you have this TODO.

```suggestion
### db.server

Create a `db.server.ts` file at the root of your `/app` directory. The [`*.server.ts` file convention](https://remix.run/docs/en/main/file-conventions/-server) ensures that Remix won't add this code into the client bundle. Add the following code snippet to connect to your Neon database:
```

content/docs/guides/remix.md
```diff
@@ -0,0 +1,189 @@
...
+export { sql };
+```
+
+```javascript
+import { neon } from '@neondatabase/serverless';
+
+const sql = neon(process.env.DATABASE_URL);
+
+export { sql };
+```
```
brookslybrand, line 98: To avoid some pain with maxing out connection limits while running the dev server I've seen some guides adopt [this strategy](https://github.com/remix-run/indie-stack/blob/main/app/db.server.ts) which leverages this [singleton implementation](https://github.com/remix-run/indie-stack/blob/main/app/singleton.server.ts)

content/docs/guides/remix.md
PaulieScanlon: That's great! Thanks @brookslybrand

content/docs/guides/remix.md
```diff
@@ -0,0 +1,189 @@
...
+
+<CodeTabs labels={["node-postgres", "postgres.js", "Neon serverless driver"]}>
+
+```javascript
+import { json } from '@remix-run/node';
+import { useLoaderData } from '@remix-run/react';
+import { pool } from '~/db.server';
+
+export const loader = async () => {
+  const client = await pool.connect();
```
brookslybrand, line 113: This is probably just my ignorance, but is there a reason you want to start the connection here vs in the `db.server.ts` file?

content/docs/guides/remix.md
```diff
@@ -0,0 +1,189 @@
...
+    console.log(response.rows[0]);
+    return json({ data: response.rows[0] });
+  } finally {
+    client.release();
+  }
+};
+
+export default function Page() {
+  const data = useLoaderData();
+}
```
brookslybrand, line 126: Totally up to you, but I'd recommend including an example where you are actually rendering something. I don't know what the data looks like exactly, but if it's just an object with a version you could do

```
export default function Page() {
  const data = useLoaderData()
  
  return (
    <h1>Version {data.version}</h1>
  )
```

content/docs/guides/remix.md
```diff
@@ -0,0 +1,189 @@
...
+```
+</CodeTabs>
+
+
+## Run the app
```
brookslybrand, line 163: I think you might actually be missing the command to run the app

content/docs/guides/remix.md
PaulieScanlon: That's a good point. This guide is the first one we've done where the database / connection are abstracted away. All our other guides show the import and connect in the one file. 

In this case I think I might make sense to move the connection to `db.server.ts`. Great suggestion!

content/docs/guides/remix.md
PaulieScanlon: Also a good point. We have added a note under "Run the app" which explains what the server output should look like. In my test apps I use...

```
export default function Page() {
  const data = useLoaderData();

  return <pre>{JSON.stringify(data, null, 2)}</pre>;
}
```

content/docs/guides/remix.md
PaulieScanlon: Ok, i've looked into this. I don't think it makes anything any clearer. I will need to check with some folks to see if there are any benefits to abstracting the pool.connect. This is what I ended up with.




```
import pg from 'pg';

const getDb = async () => {
  const pool = new pg.Pool({
    connectionString: process.env.DATABASE_URL,
    ssl: true,
  });

  const client = await pool.connect();

  return client;
};

export { getDb };
```

```
import { json } from '@remix-run/node';
import { useLoaderData } from '@remix-run/react';
import { getDb } from '~/db.server';

export const loader = async () => {
  const client = await getDb();

  try {
    const response = await client.query('SELECT version()');
    console.log(response.rows[0]);
    return json({ data: response.rows[0] });
  } finally {
    client.release();
  }
};

export default function Page() {
  const data = useLoaderData();
  return <pre>{JSON.stringify(data, null, 2)}</pre>;
}
```

content/docs/guides/remix.md
danieltprice: Makes sense to me. Let's try the quote and link. If it looks odd, we'll revise.

PR #1108, opened 2023-10-13
docs/add-run-command-next-and-astro

PaulieScanlon:
- add npm run dev command

### DIFF
--- a/content/docs/guides/nextjs.md
+++ b/content/docs/guides/nextjs.md
@@ -344,7 +344,7 @@ export const config = {
 
 ## Run the app
 
-For API handlers, and server functions you can expect to see one of the following in your terminal output:
+When you run `npm run dev` you can expect to see one of the following in your terminal output:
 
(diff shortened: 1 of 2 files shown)


content/docs/guides/nextjs.md
```diff
@@ -344,7 +344,7 @@ export const config = {
 
 ## Run the app
 
-For API handlers, and server functions you can expect to see one of the following in your terminal output:
```
PaulieScanlon, line 347: Not sure if we need to explicitly say this applies to both API Handlers and Server functions. Including that made this sentence feel a bit clunky. Watta you think @danieltprice / @bgrenon?

bgrenon: I probably wouldn't include it. I like how you've worded it. "When you run..." is nice language, it assumes the reader knows the command already, but also gets it on paper for anyone who needs the reminder.

bgrenon (approved):

PR #1115, opened 2023-10-17
docs/remix-guide-1

PaulieScanlon:
-  add remix guide

Sorry @danieltprice, I really messed up a rebase on the previous branch. I've made all the relevant changes apart from the request to add a quote and link to the header section. If we are going to do this, would we need to do it to all the framework guides and not just Remix?

dprice: I attached some unrelated minor doc updates to this PR.


danieltprice:
Hey Paul, the other approach is to revise the wording to make it your own, that way we don't need to quote. Sometimes, I'll find a few descriptions, and come up with something "new". Maybe we can do that here.

PaulieScanlon:
@danieltprice, good suggestion. OK, I'll give it a try and push the changes shortly. Thanks!

content/docs/guides/remix.md
```diff
@@ -0,0 +1,187 @@
...
+
+```shell
+DATABASE_URL=postgres://[user]:[password]@[neon_hostname]/[dbname]
+```
+
+</CodeBlock>
+
+## Configure the Postgres client
+
+There are two parts to connecting a Remix application to Neon. The first is `db.server`, Remix will ensure any code added to this file won't be included in the client bundle. The second is the route where the connection to the database will be used.
```
danieltprice, line 62:
```suggestion
There are two parts to connecting a Remix application to Neon. The first is `db.server`. Remix will ensure any code added to this file won't be included in the client bundle. The second is the route where the connection to the database will be used. 
```

content/docs/guides/remix.md
```diff
@@ -0,0 +1,187 @@
...
+
+</CodeBlock>
+
+## Configure the Postgres client
+
+There are two parts to connecting a Remix application to Neon. The first is `db.server`, Remix will ensure any code added to this file won't be included in the client bundle. The second is the route where the connection to the database will be used. 
+
+### db.server
+
+Create a `db.server.ts` file at the root of your `/app` directory add the following code snippet to connect to your Neon database:
```
danieltprice, line 66:
```suggestion
Create a `db.server.ts` file at the root of your `/app` directory, and add the following code snippet to connect to your Neon database:
```

danieltprice (approved):

PR #1122, opened 2023-10-19
disable support for free tier

bacher:
Related to https://github.com/neondatabase/company_projects/issues/154


danieltprice:
Hi Ilia, looks like you can still open a support ticket on Free. Will this be dropped?
image

bacher:
@danieltprice 
> Hi Ilia, looks like you can still open a support ticket on Free. Will this be dropped?

Console is not updated yet. I think it's better to update docs first, because otherwise it will be lie.

This is issue where we're hiding Support button: https://github.com/neondatabase/cloud/issues/7439

danieltprice (approved):

PR #1131, opened 2023-10-24
Add section about PgBouncer support for prepared statements

danieltprice:
Preview:
https://neon-next-git-dprice-pgbouncer-pgbouncer-pr-7249dc-neondatabase.vercel.app/docs/connect/connection-pooling#optimize-queries-with-pgbouncer-and-prepared-statements


content/docs/connect/connection-pooling.md
```diff
@@ -49,9 +49,58 @@ The **Connection Details** widget on the Neon **Dashboard** provides **Pooled co
...
+```sql
+PREPARE fetch_plan (TEXT) AS
+SELECT * FROM users WHERE username = $1;
+
+EXECUTE fetch_plan('alice');
+```
+
+`fetch_plan` here is the prepared statement's name, and `$1` acts as a parameter placeholder.
+
+The benefits of using prepared statements include:
```
bgrenon, line 69: Is there any caching supported? If you can save the prepared statement to cache, that can also speed things up.

content/docs/connect/connection-pooling.md
```diff
@@ -49,9 +49,58 @@ The **Connection Details** widget on the Neon **Dashboard** provides **Pooled co
...
 
+## Optimize queries with PgBouncer and prepared statements
+
+Protocol-level prepared statements are supported with Neon and PgBouncer as of the [PgBouncer 1.21.0 release](https://github.com/pgbouncer/pgbouncer/releases/tag/pgbouncer_1_21_0). Using prepared statements can help boost query performance while providing an added layer of protection against potential SQL injection attacks.
+
+### Understanding prepared statements
+
+A prepared statement in Postgres allows for the optimization of an SQL query by defining its structure once and executing it multiple times with varied parameters. Here's an SQL-level example to illustrate, but please note that direct SQL-level `PREPARE` and `EXECUTE` are not supported with PgBouncer (see [below](#use-prepared-statements-with-pgbouncer)).
+
+```sql
```
bgrenon, line 60: It seems a bit odd to provide a sample that you can't actually use. Is this because you kind of need to understand the direct SQL-level in order to correctly use the protocol-level? Or the user is familiar with the SQL-level, and so will look in the protocol library for whatever they need?

Maybe modify it to something like this, to guide the user past any weirdness:

"Here's an SQL-level example to illustrate. Note that direct SQL-level `PREPARE` and `EXECUTE` are not supported with PgBouncer (see [below](#use-prepared-statements-with-pgbouncer)), so you can't use this query from the SQL editor. It is meant to give you a clear idea of how a prepared statement works. We'll use the same variables in the protocol-level samples below, so you can see the direct SQL-level translates to different protocols."

And then should we revise the samples to use the same prepared plan name? Maybe "fetch-user" in all cases?

content/docs/connect/connection-pooling.md
```diff
@@ -49,9 +49,58 @@ The **Connection Details** widget on the Neon **Dashboard** provides **Pooled co
...
+The benefits of using prepared statements include:
+
+- **Performance**: Parsing the SQL and creating the execution plan happens just once, speeding up subsequent executions. This performance benefit would be most noticeable on databases with heavy and repeated traffic.
+- **Security**: By sending data values separately from the query, prepared statements reduce the risk of SQL injection attacks.
+
+You can learn more about prepared statements in the PostgreSQL documentation. See [PREPARE](https://www.postgresql.org/docs/current/sql-prepare.html).
+
+### Use prepared statements with PgBouncer
+
+PgBouncer supports protocol-level prepared statements only. Direct SQL-level `PREPARE` and `EXECUTE` are not supported, which means you cannot run prepared statements from an SQL client like `psql`. You must rely on PostgreSQL client libraries instead. Fortunately, most PostgreSQL client libraries support prepared statements. Here are a couple of examples showing how to use prepared statements with Javascript and Python client libraries:
```
bgrenon, line 78: Since we've already said direct SQL is not supported, maybe soften the restatement here. Something like:

"Since pgBouncer supports protocol-level prepared statements only, you must rely on PostgreSQL client libraries instead (direct SQL-level `PREPARE` and `EXECUTE` are not supported)..."

danieltprice: Updated. Thank you!

content/docs/connect/connection-pooling.md
danieltprice: Great idea, perhaps for @raoufchebri's blog as well. Modified slightly, as I am not sure I about modifying the Python example code in a way that you might not otherwise.

content/docs/connect/connection-pooling.md
danieltprice: My understanding is that the caching of prepared statement query plans would be handled by the PostgreSQL planner and executor. There's no additional user action here.

bgrenon (approved):

PR #1159, opened 2023-11-03
feat: Add Postgres docs

vannyle:
This pull request implements Algolia DocSearch for Postgres

**Aditional changes**
- removed `.html` from the Postgres links  
- adjusted the sidebar, remove `\\` symbols
- adjusted the unordered lists in posts, where there is a table of contents

Preview:
https://neon-next-git-postgres-search-neondatabase.vercel.app/docs/postgres/index


andrewgolovanov (approved):

ruf-io:
This is huge, but I think there are still some improvements we can make before shipping.

## Content

- [ ] **Can we delete the entire Preface, Tutorial, Internals, Bibliography and Book Index sections?** They don't seem relevant to Neon's needs for Postgres Docs

## Search

- [ ] **Merge Indexes, or Search both indexes at once** - It's counterintuitive that, to find certain results, I have to first toggle over to the "Postgres Docs" tab in navigation. We should have a single index where Neon docs always rank higher.

- [ ]  **Don't Index full text of Postgres Docs - Prioritize headings in rankings.** - It seems like the full text of docs are indexed, I think it's better with Algolia to only index the page titles and h1-h4s, so that if I search for something like `max` I get the actual reference for the `max` function and not random mentions of the word max.
image

## UI

- [ ] Fix or remove broken in-page anchor links - for example top paragraph here https://neon-next-git-postgres-search-neondatabase.vercel.app/docs/postgres/functions-aggregate#FUNCTIONS-AGGREGATE-TABLE
image

danieltprice:
@ruf-io Maybe this for the admonition to let users know that not everything applies? 

"You are viewing the Neon-mirrored version of the official PostgreSQL documentation. While Neon is a serverless Postgres platform built on standard Postgres, not all guidance and features detailed here may be applicable or supported by Neon."

danieltprice:
My concern with a combined search is that Neon search results get buried. If we can rank Neon results higher, as suggested, maybe that will work, but let's only proceed if that's possible.

andrewgolovanov (approved):

PR #1221, opened 2023-11-30
Update Neon CLI installation instructions

danieltprice:
Preview:
https://neon-next-git-dprice-update-cli-install-ins-133a21-neondatabase.vercel.app/docs/reference/cli-install

### DIFF
--- a/content/docs/reference/cli-install.md
+++ b/content/docs/reference/cli-install.md
@@ -34,15 +34,19 @@ brew install neonctl
 
 <TabItem>
 
-To install a binary:
+To install a [binary](https://github.com/neondatabase/neonctl/releases):
 
 - **macOS**
 
-    Download the macOS binary: 
+    Download the macOS binary:
+
+    <CodeBlock shouldWrap>
 
     ```bash
     curl -sL https://github.com/neondatabase/neonctl/releases/latest/download/neonctl-macos -o neonctl
     ```
+
+    </CodeBlock>
 
     No installation is required. Run the Neon CLI as follows:
 
@@ -52,11 +56,15 @@ To install a binary:
 
 - **Linux**
 
-    Download the Linux binary: 
+    Download the Linux binary:
+
+    <CodeBlock shouldWrap>
 
     ```bash
     curl -sL https://github.com/neondatabase/neonctl/releases/latest/download/neonctl-linux -o neonctl
     ```
+
+    </CodeBlock>
 
     No installation is required. Run the Neon CLI as follows:
 
@@ -68,9 +76,13 @@ To install a binary:
 
     Download the Windows binary:
 
+    <CodeBlock shouldWrap>
+
     ```bash
     curl -sL -O https://github.com/neondatabase/neonctl/releases/latest/download/neonctl-win.exe
     ```
+
+    </CodeBlock>
 
     No installation is required. Run the Neon CLI as follows:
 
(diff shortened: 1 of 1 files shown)


bacher:
I think it's strange that only Linux has `npm variant`. I think we should or put `npm` to all OSes, or left it only as separate Tab.

content/docs/reference/cli-install.md
```diff
@@ -7,41 +7,71 @@ updatedOn: '2023-11-08T18:42:10.264Z'
...
+```bash
+brew install neonctl
 ```
 
-Homebrew is also supported:
+</TabItem>
+
+<TabItem>
+
+Download the [neonctl-win.exe binary](https://github.com/neondatabase/neonctl/releases/tag/v1.24.0).
```
bacher, line 37: I think it would be better to move word binary outside of link:

image

Link on Linux tab:

image

danieltprice:
@bayandin As a Linux user, which installation methods do you think we should document for Linux?

bayandin:
> @bayandin As a Linux user, which installation methods do you think we should document for Linux?

I'm not a real Linux user, but I would imagine that downloading binary is the most popular way on it. 
@duskpoet, who is a real Linux user, what do you think?

content/docs/reference/cli-install.md
```diff
@@ -7,41 +7,71 @@ updatedOn: '2023-11-08T18:42:10.264Z'
...
 
-   ```shell
-   npm -v
-   ```
 
-  If you need to install `Node.js` or `npm`, refer to instructions on the [official nodejs page](https://nodejs.org) or use the [Node version manager](https://github.com/nvm-sh/nvm).
+</TabItem>
 
-### Install
+<TabItem>
```
duskpoet, line 25: we have binary for macOS too :)

duskpoet (approved):

content/docs/reference/cli-install.md
danieltprice: Added. TY.

danieltprice:
@duskpoet Was this along the lines of the bash snippet for download the latest binary:

# Get the latest release tag from GitHub API
latest_tag=$(curl -s https://api.github.com/repos/neondatabase/neonctl/releases/latest | awk -F '"' '/tag_name/{print $4}')

# Download the latest binary using the latest tag
curl -sL "https://github.com/neondatabase/neonctl/releases/download/${latest_tag}/neonctl-linux" -o neonctl

# Make the binary executable
chmod +x neonctl

If so, what are your thoughts on adding the snippet to the docs? Were you thinking to just add it as a code block?

duskpoet:
> @duskpoet Was this along the lines of the bash snippet you were thinking of for download the latest binary:
> 
> * Get the latest release tag from GitHub API
>   latest_tag=$(curl -s https://api.github.com/repos/neondatabase/neonctl/releases/latest | awk -F '"' '/tag_name/{print $4}')
> * Download the latest binary using the latest tag
>   curl -sL "[https://github.com/neondatabase/neonctl/releases/download/${latest_tag}/neonctl-linux](https://github.com/neondatabase/neonctl/releases/download/$%7Blatest_tag%7D/neonctl-linux)" -o neonctl
> * Make the binary executable
>   chmod +x neonctl
> 
> If so, what are your thoughts on adding the snippet to the docs? Were you thinking to just add it as a code block?

Yea, something like that! Let's add it

bayandin:
> * Get the latest release tag from GitHub API
>   latest_tag=$(curl -s https://api.github.com/repos/neondatabase/neonctl/releases/latest | awk -F '"' '/tag_name/{print $4}')
> * Download the latest binary using the latest tag
>   curl -sL "[https://github.com/neondatabase/neonctl/releases/download/${latest_tag}/neonctl-linux](https://github.com/neondatabase/neonctl/releases/download/$%7Blatest_tag%7D/neonctl-linux)" -o neonctl

There's no need to parse tags:

```bash
curl -sL https://github.com/neondatabase/neonctl/releases/latest/download/neonctl-linux -o neonctl
```

PR #1228, opened 2023-12-01
Update cli-completion.md

danieltprice:
Fix wording / typos.

### DIFF
--- a/content/docs/reference/cli-completion.md
+++ b/content/docs/reference/cli-completion.md
@@ -19,7 +19,11 @@ This command generates a completion script for the `neonctl` command-line interf
 neonctl completion
 ```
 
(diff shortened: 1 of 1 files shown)


content/docs/reference/cli-completion.md
```diff
@@ -19,7 +19,11 @@ This command generates a completion script for the `neonctl` command-line interf
 neonctl completion
 ```
 
-The command outputs the following completion script similar to the following with installation instructions. **Use the completeion scription that is output to your terminal or command window**, as the script may differ depending on your operating environment. 
+The command outputs a completion script similar to the one shown below.
+
+<Admonition type="important">
+Use the completion scription that is output to your terminal or command window, as the script may differ depending on your operating environment.
```
bgrenon, line 25:
```suggestion
Use the completion script that is output to your terminal or command window, as the script may differ depending on your operating environment.
```

PR #1235, opened 2023-12-04
Logical data size updates

danieltprice:
Logical data size updates. Need to distinguish from synthetic storage size.


danieltprice:
Opened a new one. This one is corrupted.

PR #1242, opened 2023-12-05
Dprice 12 04 storage release

danieltprice:


knizhnik (approved):
I am not sure if this PR is the right place for complaints, but optimization of stopping null images is not yet completed: https://github.com/neondatabase/neon/pull/5979 is not yet reviews and merged. And without it WAL size produced by pg16 is the same as of pg15, but storage size is still 2x times larger.

conradludgate (approved):

bgrenon (approved):

PR #1310, opened 2024-01-10
SignUp shared component

bgrenon:
A "SignUp" shared component intended to be used in the postgres functions pages.

Once text and approach is approved, let's see about getting @k8isdead to style this into a card or some other visual that nudges the user towards trying out the postgres content described on the page in Neon, without annoying the user or detracting/distracting too much from the point of the page (describe a postgres function or extension for a general postgres audience).

Example:
image


danieltprice:
We could also use this on the extensions and types pages. Instead of a card, which is half-page width, I think it might look better if it were full-page width.

danieltprice (approved):
Text looks good to me.

danieltprice:
@ruf-io Assuming that users land in our docs via the Frindle content, this component would serve as a CTA to sign up try the examples in Neon.

danieltprice (approved):

PR #1313, opened 2024-01-10
Update sidebar.yaml

bgrenon:
Updated for a new Postgres guides section. Includes pointer in toc for the [json_build_object page](https://neon.tech/docs/functions/json_build_object) that is already published.


danieltprice:
The sidebar json_build_function link is broken. It looks like this branch does not yet include the json_build_function.md file. You can fix this by merging changes from main into this branch, I think.

danieltprice (approved):

PR #1327, opened 2024-01-16
Added download mention to overview page

bgrenon:


danieltprice:
If it makes sense, I would add something to our security docs about SOC 3 and what it is since we describe SOC 2 Type 1 and 2. @areyou1or0 ?

danieltprice (approved):

bgrenon:
> If it makes sense, I would add something to our security docs about SOC 3 and what it is since we describe SOC 2 Type 1 and 2. @areyou1or0 ?

@areyou1or0 As you described it [here](https://neondb.slack.com/archives/C027XFH0448/p1705394647818929), SOC 2 is not available to download, while SOC 3 is a publically available version of SOC 2 that users can download. And the behaviour matches this description in the Trust Center. I've revised the text accordingly. Can you please review?

PR #1330, opened 2024-01-16
Drop and recreate pg_stat_statements for new privilege

danieltprice:
Drop and recreate pg_stat_statements for new privilege


Bodobolero (approved):
thanks, looks good

PR #1356, opened 2024-01-23
Fix broken links

danieltprice:
Fix a few broken links


bgrenon (approved):
Woops! I should have caught those missing slashes. Thank you!

PR #1358, opened 2024-01-23
Add execution time queries to pg_stat_statements topic

danieltprice:


brendan-stephens:
For `Look up queries by execution time` it might be better to specify `queries by last execution time` or `queries run within a time interval` to help disseminate `execution time` (how long something took to run) vs (when it was last run). 

On the second query, when user was added, it changed the sort order.
That query is designed to find the most frequently executed, ordered by the number of calls. 

Everything else is golden.

brendan-stephens:
I forgot to show the difference...
We can either change the `order by` to the ordinal position of the selection, or we can change the query... 

```sql
SELECT
  (total_exec_time / 1000 / 60) as total_min,
  mean_exec_time as avg_ms,
  userid,
  calls,
  query
FROM pg_stat_statements
ORDER BY 3 DESC
LIMIT 10;
```

But this might look better...
User, query, number of times it has been called, average runtime. 

```sql
SELECT
  userid,
  query,
  calls,
  (total_exec_time / 1000 / 60) as total_min,
  mean_exec_time as avg_ms
FROM pg_stat_statements
ORDER BY 3 DESC
LIMIT 10;
```

danieltprice:
Thank you @brendan-stephens 
Can you take another look? I revised the headings, simplified the query descriptions, and replaced the second query with the better looking one that you suggested.

content/docs/extensions/pg_stat_statements.md
```diff
@@ -195,6 +195,46 @@ This query returns results similar to the following:
...
+<Admonition type="note">
+In Neon, this query can help identify SQL queries that are waking up an idle compute endpoint.
+</Admonition>
+
+```sql
+SELECT
+  userid,
+  query,
+  calls,
+  total_time,
```
bgrenon, line 211: Should this be "total_exec_time"? I get: column "total_time" does not exist .

PR #1363, opened 2024-01-24
Docs: Remove underscores from anchors

ruf-io:
I noticed that our anchor link creation algo removes underscores, but some of our anchor links were including them. This change ensures that internal anchor links on headings with underscores take a user to the right position in page.


danieltprice (approved):
Thank you!

PR #1430, opened 2024-02-10
pg_prewarm extension

danieltprice:
Another extension guide:
https://neon-next-git-dprice-pg-prewarm-neondatabase.vercel.app/docs/extensions/pg_prewarm


content/docs/extensions/pg_prewarm.md
```diff
@@ -0,0 +1,241 @@
+---
+title: The pg_prewarm extension
+subtitle: load data into Postgres shared buffers with the pg_prewarm extension
```
bgrenon, line 3:
```suggestion
subtitle: Load data into Postgres shared buffers with the pg_prewarm extension
```

content/docs/extensions/pg_prewarm.md
```diff
@@ -0,0 +1,241 @@
...
+    ```sql
+    SELECT pg_prewarm('public.t_test') AS blocks_loaded;
+    ```
+
+    This will output the number of blocks that were loaded:
+
+    ```sql
+    blocks_loaded 
+    ---------------
+            4480
```
bgrenon, line 114: I got 4425 for some reason

bgrenon: Ah, you explain some possible reasons below. :)

content/docs/extensions/pg_prewarm.md
```diff
@@ -0,0 +1,241 @@
...
+
+    ```sql
+    EXPLAIN ANALYZE SELECT * FROM tbl_transactions_1;
+    ```
+
+    ```sql
+    EXPLAIN ANALYZE SELECT * FROM tbl_transactions_2;
+    ```
+
+    The execution time for the prewarmed table should be significantly lower than for the table that has not been prewarmed, as shown here:
```
bgrenon, line 213: I'm not sure what counts as "significantly lower" but this is what I got:

Seq Scan on tbl_transactions_1  (cost=0.00..69608.21 rows=4252321 width=18) (actual time=0.002..188.812 rows=4252321 loops=1)
Planning Time: 0.119 ms
Execution Time: 282.806 ms

Seq Scan on tbl_transactions_2  (cost=0.00..69608.21 rows=4252321 width=18) (actual time=0.003..203.255 rows=4252321 loops=1)
Planning Time: 0.129 ms
Execution Time: 293.606 ms

content/docs/extensions/pg_prewarm.md
```diff
@@ -0,0 +1,241 @@
...
+    ---------------------------------------------------------------------------------------------------------------------------------
+    Seq Scan on tbl_transactions_2  (cost=0.00..69608.21 rows=4252321 width=18) (actual time=2.251..11859.232 rows=4252321 loops=1)
+    Planning Time: 0.216 ms
+    Execution Time: 11994.066 ms
+    (3 rows)
+    ```
+
+## Conclusion
+
+Prewarming your table data and indexes can help improve read performance, especially after a database restart or for indexes that are not frequently used but will be involved in upcoming heavy read operations. However, it's important to use this feature cautiously, especially on systems with limited memory, to avoid potential negative impacts on overall performance.
```
bgrenon, line 235: Are there any cost/compute implications when using this with Neon?

content/docs/extensions/pg_prewarm.md
```diff
@@ -0,0 +1,241 @@
...
+
+    You can now compare this value with the size of your table.
+
+    ```sql
+     loaded_data_bytes 
+    -------------------
+            36700160
+    ```
+
+    The values for the size of the table and the size of the data loaded into the buffer cache as shown in the example above match exactly, which is an ideal scenario. However, there are cases where these values might not match, indicating that not all the data was loaded into the buffer cache; for example:
```
bgrenon, line 146: I found this section a bit dense. I don't know if it can be simplified; it may be the nature of the content. But something to consider.

bgrenon (approved):
Great doc! I did not test the index queries, but did test everything else.

content/docs/extensions/pg_prewarm.md
danieltprice: Checked again to make sure.

content/docs/extensions/pg_prewarm.md
danieltprice: I think stopping the compute to release memory might have been the difference here.

content/docs/extensions/pg_prewarm.md
danieltprice: TY. I shortened it.

PR #1474, opened 2024-02-23
Update billing-sample.md

danieltprice:


bgrenon (approved):
Nice add about using the extra allowance to increase the size or enable autoscaling!

PR #1481, opened 2024-02-27
Add Cookie Banner

ruf-io:
This PR does three things:

1. Adds Zaraz - this is a cloudflare alternative to google tag manager currently hosted on neondata.cloud - need to transfer to an official neon subdomain account before merging.
2. Removes Google Tag Manager - Zaraz can handle forwarding to GA4 and other tags when consent is applied
3. Removes Segment - Zaraz can handle forwarding to Segment when consent is applied.


src/app/layout.jsx
```diff
@@ -18,13 +18,7 @@ export const preferredRegion = 'edge';
...
   <html lang="en">
     <head>
-      {process.env.NODE_ENV === 'production' && (
-        <Script id="google-tag-manager" strategy="afterInteractive">
-          {`
-        (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start': new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0], j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src= 'https://www.googletagmanager.com/gtm.js?id='+i+dl+'';f.parentNode.insertBefore(j,f); })(window,document,'script','dataLayer', 'GTM-MJLTK6F');
-      `}
-        </Script>
-      )}
+      <Script strategy="afterInteractive" src="https://neondata.cloud/cb.js" />
```
vannyle, line 21: Hi @ruf-io, I've noticed that we still use the function `sendGtagEvents`, but the gtag script is removed for some reason. I'd suggest returning it according to the guide: https://developers.google.com/tag-platform/tag-manager/datalayer

ruf-io: The new cookie consent manager script from Zaraz is operating in [GTM Compatibility mode](https://developers.cloudflare.com/zaraz/advanced/datalayer-compatibility/#:~:text=Cloudflare%20Zaraz%20offers%20backwards%20compatibility,will%20automatically%20collect%20your%20events.) so (at least in my testing) we don't even need to load the GTM stuff any more. Zaraz watches changes to datalayer and handles the required call server-side.

So far in testing it works, seems like a nice upgrade where we can run fewer third-party scripts on the site.

PR #1533, opened 2024-03-14
Mention that Neon CLI Linux binary does not support ARM64

danieltprice:


duskpoet (commented):
I've added the arm64 binary

danieltprice:
Thank you @duskpoet

PR #1537, opened 2024-03-15
Update Django content

danieltprice:
- Add note about using a recent driver version
- Use the same connection snippet we provide in the Neon console
- Reference Evan's Django blog post and sample app
Preview:
https://neon-next-git-dprice-update-django-content-neondatabase.vercel.app/docs/guides/django

### DIFF
--- a/content/docs/guides/django.md
+++ b/content/docs/guides/django.md
@@ -82,14 +82,12 @@ The version number for `libpq` is presented in a different format, for example, 
 
 ## Django application blog post and sample application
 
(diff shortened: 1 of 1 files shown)


evanshortiss (approved):
LGTM. Should we show an example of passing the `endpoint` parameter for older drivers? Or is that something we'd rather not encourage?

danieltprice:
> LGTM. Should we show an example of passing the `endpoint` parameter for older drivers? Or is that something we'd rather not encourage?

My thought is that we should encourage users to update their driver to avoid things breaking later when they do happen to update their driver.

PR #1568, opened 2024-03-26
Mention possible operation status values

danieltprice:
Mention Operation status values in the Operation docs

+ minor unrelated update to import docs to fix an error.


danieltprice:
> Would be great to also add examples of what each status means and what to do in case of errors.

danieltprice:
> > Would be great to also add examples of what each status means and what to do in case of errors.

Thank you. I'll do another PR. Let's get this one out for now.

PR #1581, opened 2024-04-02
Fix minor documentation typo/mismatch

diego-lipinski-de-castro:


danieltprice (approved):

danieltprice:
Thank you @diego-lipinski-de-castro

PR #1598, opened 2024-04-08
Mention configuring client & apps to reconnect automatically

danieltprice:
Mention configuring client & apps to reconnect automatically


content/docs/manage/endpoints.md
```diff
@@ -204,7 +206,7 @@ The maximum **Suspend compute after a period of inactivity** setting is 7 days.
 It is sometimes necessary to restart a compute endpoint. For example, if you upgrade to a paid plan account, you may want to restart your compute endpoint to immediately apply your upgraded limits.
 
 <Admonition type="important">
-Please be aware that restarting a compute endpoint interrupts any connections currently using the compute endpoint.
+Please be aware that restarting a compute endpoint interrupts any connections currently using the compute endpoint. To avoid prolonged interruptions resulting from compute restarts, we recommend configuring your clients and applications to reconnect automatically in case of a dropped connection.
 </Admonition>
 
 You can restart a compute endpoint using one of the following methods:
```
zaynetro, line 212: We are now able to restart a compute endpoint in a single API call since yesterday's deployment: https://console.neon.tech/api-ref/v2#/Endpoint/restartProjectEndpoint

You can update the steps below now or later. API docs on the website are not yet synced apparently.

zaynetro (approved):

content/docs/manage/endpoints.md
danieltprice: @zaynetro For some reason, the Restart endpoint is not showing in our API docs (rendered/hosted by [Readme](https://readme.com/)). Readme is less permissive than Swagger. Can you take a look at the Restrt endpoint API and compare it to the Start endpoint API? One thing I noticed is that the parameters for Restart are defined outside of the post method for Restart, whereas parameters are defined within the post method for Start. I am not sure if that has anything to do with it, but it's the only substantive difference I see.

The new endpoint is present in the public API spec that is rendered by Readme:
https://dfv3qgd2ykmrx.cloudfront.net/api_spec/release/v2.json
And other API changes in the same console release do show up in the API docs as expected.

What GPT thinks:
image

zaynetro: `/restart` copies `/suspend` endpoint so the definition should be fine.

Appears that now it is synced https://api-docs.neon.tech/reference/restartprojectendpoint

I have no idea when that syncing happens.

danieltprice: comment above addressed.

PR #1599, opened 2024-04-09
Update possible status values list

danieltprice:
We'll define these values as a next step. Capturing them for now.

### DIFF
--- a/content/docs/manage/operations.md
+++ b/content/docs/manage/operations.md
@@ -38,7 +38,17 @@ Operation details include:
 - **Duration**: The duration of the operation.
 - **Date**: The date and time the operation occurred.
 
-Possible **Status** values are `OK`, `Scheduling`, `In progress`, and `Error`.
+Possible **Status** values are:
+
+- `cancelled`
+- `cancellation_pending`
+- `error`
+- `failed`
+- `finished`
+- `running`
+- `scheduling`
+- `scheduling`
+- `skipped`
 
 </TabItem>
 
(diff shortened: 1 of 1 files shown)


content/docs/manage/operations.md
```diff
@@ -38,7 +38,17 @@ Operation details include:
...
+Possible **Status** values are:
+
+- `cancelled`
+- `cancellation_pending`
+- `error`
+- `failed`
+- `finished`
+- `running`
+- `scheduling`
+- `scheduling`
```
ololobus, line 50: Duplicated

content/docs/manage/operations.md
```diff
@@ -314,6 +324,18 @@ curl 'https://console.neon.tech/api/v2/projects/autumn-disk-484331/operations/05
...
+Possible operation `status` values include:
+
+- `cancelled`
+- `cancellation_pending`
+- `error`
+- `failed`
+- `finished`
+- `running`
+- `scheduling`
+- `scheduling`
```
ololobus, line 336: And here

ololobus (approved):
LGTM, but I wonder, should we expose them in docs now? They are not used yet, so can also create a confusion

content/docs/manage/operations.md
danieltprice: fixed

danieltprice:
> LGTM, but I wonder, should we expose them in docs now? They are not used yet, so can also create a confusion

We'll hold off on publishing for now. I'll take a closer look next week.

danieltprice:
We'll wait until they are exposed in API before publishing.

PR #1616, opened 2024-04-13
[ON HOLD] add doc for extension postgresql-anonymizer

ananis25:
add doc for extension postgresql-anonymizer


danieltprice:
We're going a different direction.

PR #1633, opened 2024-04-19
Serverless driver ws bufferutil requirement

danieltprice:


content/docs/guides/prisma.md
```diff
@@ -178,6 +178,14 @@ const prisma = new PrismaClient({ adapter })
 
 You can now use Prisma Client as you normally would with full type-safety. Prisma Migrate, introspection, and Prisma Studio will continue working as before, using the Neon connection string defined by the `DATABASE_URL` variable in your `schema.prisma` file.
 
+<Admonition type="note">
+If you encounter a `TypeError: bufferUtil.mask is not a function` error when building your application, this is likely due to a missing Websocket implementation required to use the `Client` or `Pool` constructs with the Neon serverless driver. You can address this requirement by installing the `bufferutil` package:
```
evanshortiss, line 182: `this is likely due to a missing Websocket implementation`

I think it's more accurate to say "a missing dependency that the `ws` module requires when using Client and Pool constructs`.

I'd need to investigate further, but **I think** that `bufferutil` is only necessary if you provide the `ws` constructor to `neonConfig`. AFAIK, it also provides a performance boost and that's why `ws` uses it. I'm not sure why `ws` doesn't just install it alongside itself. Also, I'm trying to recall if it works locally without `bufferutil` but breaks when deployed.

evanshortiss:
Looks good. I'd just change the wording as suggested.

content/docs/guides/prisma.md
danieltprice: @evanshortiss Adjusted the text as suggested. Should we publish what we have for now and clarify when learn more?

evanshortiss (approved):

PR #1655, opened 2024-05-03
feat: redirect known users

ruf-io:
This is testing UX of redirecting known (have `neon_login_indicator`) users directly to console.

> [!TIP]
> To make yourself a "known user" in the [Vercel preview](https://vercel.live/open-feedback/neon-next-git-ruf-io-logged-in-redirect-neondatabase.vercel.app?via=pr-comment-visit-preview-link&passThrough=1), open up web console and run `document.cookie = 'neon_login_indicator'`

## How it works

1. Redirect from `neon.tech` to `console.neon.tech` for anyone who has the `neon_login_indicator` cookie set (this is a cookie that is set when a user logs in and cleared when they log out)  https://github.com/neondatabase/website/blob/1fc1ab6b685d5451b4a9220a395466544f015ae8/src/app/(home-page)/page.jsx#L20-L32

2. Create a new clone of the homepage at neon.tech/home that is blocked from search engine indexes https://github.com/neondatabase/website/blob/da8b9b27455beb830c43dfbe5283a31187af2786/src/app/home/page.jsx#L1-L37

3. For visitors that have `neon_login_indicator`, replace the "Sign Up" and "Log In" buttons with "Go to Console" 
image


clarkbw:
I like this. I think the main thing I notice is that docs, blog, changelog links are down on the left. If I was going to `neon.tech` for a blog post I don't think I'd look there. When I compare to Vercel the top bar has those links as well as the bottom row. I don't think we need to move quickly to make a change to what they have but it's the first thing I noticed.

americano98:
Hi @ruf-io,

The PR looks great! Could we refactor step 3 slightly to utilize the constant LINKS.home? I believe there are two potential options (or a combination of both):

1. Add the property home: getHomepageRoute() to constant.js.
2. Implement a Next.js [redirect](https://nextjs.org/docs/app/api-reference/functions/redirect) from `/` to `/home` in [page.jsx](https://github.com/neondatabase/website/blob/main/src/app/page.jsx) when a cookie is present.

I believe these changes would not only help prevent users with cookies from accessing the `/` page but also simplify the management of home page links throughout our project.

ruf-io:
@americano98 Want to take over the PR and make the necessary changes?

Just hold off on merging. I need to get proper buy-in internally first.

americano98:
@ruf-io I made some changes. Now, the workflow is as follows:

When a user visits the `/` page, if they have a cookie, we check the headers to determine where the user came from. If it's an internal transition, we redirect them to `/home`. If it's an external transition, we redirect them to `console.neon.tech`.

This way, we can keep all links pointing to the index as `/`. And thanks to server actions, we will generate the UI header on the server side

PR #1704, opened 2024-05-21
Add event tracking for button clicks site-wide

ruf-io:
Please review this PR closely, I don't want to break button clicks in any way.

This adds event tracking to _every_ <Button> on the site, the structure of the event is:
```
track("Button Clicked" {"style": "<theme>", "tag_name": "<optional extra tag to help disambiguate buttons", "text": "<text of button>"})
```

For example the homepage CTA is 
```
track("Button Clicked" {"style": "primary", "tag_name": "Hero", "text": "Get Started"});
```

FYI I've also switched from the legacy compatibility-mode usage of `window.datalayer.push` to the better-supported Zaraz-native, window.zaraz.track https://developers.cloudflare.com/zaraz/web-api/track/

### DIFF
--- a/src/components/pages/home/get-started/get-started.jsx
+++ b/src/components/pages/home/get-started/get-started.jsx
@@ -37,6 +37,7 @@ const GetStarted = () => (
         theme="primary"
         to={LINKS.signup}
         target="_blank"
+        tag_name="Footer"
       >
         Get Started
       </Button>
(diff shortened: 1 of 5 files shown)


americano98:
@ruf-io sorry for pushing to this branch by mistake

ruf-io:
@americano98 any issues with converting <Button> to "use client" in this PR?

https://github.com/neondatabase/website/blob/258bbe6e0a303ff702cd9cdf7cb3d855efb1a178/src/components/shared/button/button.jsx#L1-L8

americano98:
@ruf-io  no issues, this should not affect the site's functionality. The only thing is that we might be able to split the component in the future so that it is client-side only when necessary, but for now, we can use this solution

PR #1756, opened 2024-05-30
docs: recommend users to use the Neon CLI to update pg_bouncer settings

rishi-raj-jain:


YanicNeon (changes requested):
This is incorrect.

First, API and CLI are two different things.
But most important, while it's technically feasible to specify both a pg_settings and a pgbouncer_settings object in the API call, it's not possible to change neither max_connections nor default_pool_size. In both cases, an error is thrown reporting that this settings cannot be changed.
Please cancel this PR, the previous doc version is correct.

Thanks!

rishi-raj-jain:
Noted, the PR is already cancelled.

PR #1763, opened 2024-05-31
docs: add asynchronous python guide

rishi-raj-jain:
As requested by @petuhovskiy, it's a good to have an async guide for connecting to Postgres. I'm using [asyncpg](https://github.com/MagicStack/asyncpg) for the same.

### DIFF
--- a/content/docs/guides/python.md
+++ b/content/docs/guides/python.md
@@ -69,6 +69,8 @@ DATABASE_URL=postgres://[user]:[password]@[neon_hostname]/[dbname]?sslmode=requi
 
 Add a `neon-connect.py` file to your project's root directory and add the following code. The script connects to your Neon database and retrieves the current time and Postgres version.
 
+<CodeTabs labels={["synchronous", "asynchronous"]}>
+
 ```python
 import os
 import psycopg2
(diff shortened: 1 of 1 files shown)


content/docs/guides/python.md
```diff
@@ -102,6 +104,39 @@ print('Current time:', time)
...
+
+async def main():
+   # Load .env file
+   load_dotenv()
+
+   # Get the connection string from the environment variable
+   connection_string = os.getenv('DATABASE_URL')
+
+   # Connect to the Postgres database
+   conn = await asyncpg.connect(connection_string)
```
petuhovskiy, line 121: Not sure if we should do it, but we can recommend using pools instead of a single `conn`, since it's better for server applications.

https://magicstack.github.io/asyncpg/current/usage.html#connection-pools

rishi-raj-jain: I think we may update both in this case?

petuhovskiy: Sure, let's update both.

rishi-raj-jain: Sounds good!

petuhovskiy (approved):

PR #1785, opened 2024-06-05
Update nextjs.md

Boby900:
when I was not adding the pooler flag in the production I was getting an error stating that no database connection string was provided even though I had added the database connection URL but after adding this flag it's working fine.

Screenshot (535)


Boby900:
Appreciate you reviewing the PR, @rishi-raj-jain 

I've created a minimal reproduction of the issue. You can find it [here](https://github.com/Boby900/yogurt). 

Additionally, I have deployed it to production. You can see it in the Deployments tab.

danieltprice (approved):

PR #1799, opened 2024-06-06
docs: restart compute for latest updates

danieltprice:
- Add notes about restarting your compute for the latest updates if autosuspend is disabled
- Improve compute restart instructions. Provide cURL example and information about where to find the required parameters


YanicNeon (approved):
Looks good to me. Thanks !

PR #1808, opened 2024-06-10
docs: mention pg.new in the docs

danieltprice:
Mention pg.new in the docs.

### DIFF
--- a/content/docs/manage/projects.md
+++ b/content/docs/manage/projects.md
@@ -68,6 +68,10 @@ To create a Neon project:
 
 After creating a project, you are presented with a dialog that provides your connection details for a ready-to-use `neondb` database. The connection details include your password.
 
+<Admonition type="tip">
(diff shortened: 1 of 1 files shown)


content/docs/manage/projects.md
```diff
@@ -68,6 +68,10 @@ To create a Neon project:
 
 After creating a project, you are presented with a dialog that provides your connection details for a ready-to-use `neondb` database. The connection details include your password.
 
+<Admonition type="tip">
+Similar to **docs.new** for instantly creating Google Docs or **repo.new** for adding new GitHub repositories, you can use [pg.new](https://pg.new) to create a new Neon Postgres project. Simply visit [pg.new](https://pg.new) and you'll be taken straight to the **Create project** page where you can create your new project.
```
rishi-raj-jain, line 72: Postgres project -> Postgres database?

danieltprice: It actually creates a new project. I think we say "database" when we marketed it, which is also true (you get a database with a project), but this addition is in the context of the "creating projects" in the docs, so I think it's ok.

rishi-raj-jain: aight sounds like all good

rishi-raj-jain (approved):

PR #1813, opened 2024-06-11
Update wording of the pooler 'important' box

MMeent:
IMV, this reduces confusion regarding the number of concurrent active connections that we support, and clarifies limits that will always apply even when you use the pooler.


danieltprice (approved):

danieltprice (approved):

content/docs/connect/connection-pooling.md
```diff
@@ -63,7 +63,9 @@ Even with the largest compute size, the `max_connections` limit may not be suffi
...
+Connection pooling supports up to 10,000 concurrent connections. However, these connections are not a magic bullet: These connections are pooled and take shared ownership of a pool of connections to PostgreSQL. This pool is shared between all connections that connect with the same user credentials to the same database, and share a total of up to [64 connections](#neon-pgbouncer-configuration-settings). This means any user can only have 64 concurrently active transactions in a single database when connected through the pooler endpoint.
```
rishi-raj-jain, line 66: > [64 connections](#neon-pgbouncer-configuration-settings). This means any user can only have 64 concurrently active transactions in a single database when connected through the pooler endpoint.

in my limited knowledge, a transaction can have any X number of operations in it to be executed in sequence, while a connection may or may not have a sequence of operations. would it make sense to update the wording to stick to either one?

basically, total of up to 64 connections or transactions?

content/docs/connect/connection-pooling.md
```diff
@@ -63,7 +63,9 @@ Even with the largest compute size, the `max_connections` limit may not be suffi
...
+Connection pooling supports up to 10,000 concurrent connections. However, these connections are not a magic bullet: These connections are pooled and take shared ownership of a pool of connections to PostgreSQL. This pool is shared between all connections that connect with the same user credentials to the same database, and share a total of up to [64 connections](#neon-pgbouncer-configuration-settings). This means any user can only have 64 concurrently active transactions in a single database when connected through the pooler endpoint.  
+Additionally, because PgBouncer still needs to connect to PostgreSQL, the number of active concurrent connections is further limited by `max_connections`: if the `max_connections` parameter of Postgres is 112 (i.e. the endpoint was started with 0.25 CU), you can still hit PostgreSQL's own connection limit with only 2 roles.
```
rishi-raj-jain, line 67: If the total limit here is max_connections regardless of Pooling, per min 8 CU: https://neon.tech/docs/connect/connection-pooling#connection-limits-without-connection-pooling it is upto 3604 connections.

But I know what you mean, the client can still open 10K - the Pgbouncer can open upto max_connections with pg. can we make that more clear here that here we're talking about pgbouncer to pg (a system inside of Neon?)?

content/docs/connect/connection-pooling.md
YanicNeon: 64 transactions. Our pgbouncer runs in transaction mode.
https://neon.tech/docs/connect/connection-pooling#neon-pgbouncer-configuration-settings

rishi-raj-jain: How do we communicate it simply to the users what'd 64 transactions mean?  It's common to get lost into the details.

The current/PR'd wording fails to communicate that it's about internals of Neon itself: PgBouncer - Postgres (within Neon) which user does not have access to.

Think of it as a user who comes in with the expectation of 10K connections, pulls up this guide and then sees "For a single user connecting to a single database, the maximum number of concurrent connections is 64." which in fact is not correct as this is between PgBouncer and Postgres.

Also, while it seems great that we can open upto 10K connections, can we also document the concurrent limit of starting those connections?

content/docs/connect/connection-pooling.md
```diff
@@ -60,12 +60,22 @@ Even with the largest compute size, the `max_connections` limit may not be suffi
...
+You will not be able to get interactive results from all 10,000 connections at the same time. Connections to the pooler endpoint still consume some connections on the main endpoint: PgBouncer forwards operations from the user's connections through its own pool of connnections to Postgres, and adaptively adds more connections to PostgreSQL as and when needed by more concurrently active user connections. The 10,000 connection limit is therefore most useful for "serverless" applications and application-side connection pools that have many open connections, but infrequent and/or short transactions.
```
rishi-raj-jain, line 74: What's the limit of having concurrent connections? Should we document that?

rishi-raj-jain, line 74: > but infrequent and/or short transactions

- what's a "short" transaction?
- each connection that executes a query, and takes greater than `query_wait_timeout` time is a "long" transaction?

rishi-raj-jain: > That depends on the application, but generally I'd transactions short when the duration of the whole transaction is within one order of magnitude of the RTT to the application.

Shall we update the wording? Because this gets caught in a loop. So a hypothetical case that's bound to come:

- a user assumes/thinks that their transaction is short
- they report
- they'd be let known that their transaction was instead long
- they come back - ask for what's a long transaction?
- well, that depends and is specific to your case, which in this case it was it

rishi-raj-jain: > I'd say so, yes, but that depends on query_wait_timeout. You can configure that setting down to 1ms without issues (other than that most significant queries will fail at that point).

Let's definitely including the `query_wait_timeout` as the limit for long or short transaction?

> You can configure that setting down to 1ms without issues (other than that most significant queries will fail at that point).

Would we want to document this?

acervantes23: Let's keep the changes as proposed and for "short" transactions let's just link to our hosted postgres docs so people can read more if need be: https://neon.tech/docs/postgres/tutorial-transactions .

rishi-raj-jain:
@MMeent 

I see that you created https://github.com/neondatabase/website/issues/1884. Looks like it was by mistake?

rishi-raj-jain:
> @rishi-raj-jain that was deliberate, after Atli's comments in this thread over here: https://neondb.slack.com/archives/C03ETHV2KD1/p1719345912362679?thread_ts=1719324029.899139&cid=C03ETHV2KD1

I see, thanks for clearing that up.

PR #1837, opened 2024-06-14
Update hero.jsx

carlotas19:
Add new features to hero in substitution to read replicas / IP allow


ruf-io:
@carlotas19 I know this isn't part of your changes, but can you see if we can get the Scale desc to two lines 
image

danieltprice (approved):

PR #1840, opened 2024-06-17
docs: modify cli page to show install command only

danieltprice:
We'd like to show only the CLI installation command, which was suggested by @m-abdelwahab (mockup shown below)
image

The change in the PR fails to shrink the height of the tabbed codeblock.


rishi-raj-jain (approved):

saimonkat:
Hi @danieltprice. Unfirtunately we don't have a shrink for the height of the tabbed codeblock in case of the same height for different tabs. So we have it's value hardcoced in `code-tabs-navigation.jsx` file and I just fixed it

image

danieltprice:
Thank you @saimonkat

PR #1920, opened 2024-07-09
feat: docs layout and navigation update

saimonkat:
This PR updates docs layout and sidebar structure

image

[Preview](https://neon-next-git-docs-style-update-neondatabase.vercel.app/docs/introduction)


danieltprice:
This is looking great. We'll do a thorough review, but a couple of items I noticed:

1. If you click an item further down in the nav, the Back to home link is hidden when the secondary nav opens:
image

2. Browser back button is problematic. Try using the nav to go to a secondary nav. Then use the browser back button.

danieltprice:
This link should take you back to the Neon docs. "Back to Neon docs"
image

danieltprice:
If you click on a link to a page in the docs, it should take you to that document in the laft nav. Here, I navigated from ampage to the pg_stat_statements doc, but the nav did not change:

image
-+

danieltprice:
This is a change from what we asked for previously. Sorry.
Every item in the primary nav should have a secondary nav in order to have consistent behavior, even if it is a single page, such as the Support page.

bgrenon:
This is looking great!

Can we: swap the "Changelog" item in the top 4 navbar links for "Connect"? We'll need a new icon for Connect.

We want to surface this content. we want to surface that. We can  move "Changelog" down into the unnamed group at the bottom of the navbar, along with "Community".

bgrenon:
@saimonkat at 100% browser zoom on my MacBook, the right sidebar navigation is floated beneath the body content.

Can the layout be tweaked to preserve the sidebar for this typical screen size?
image

saimonkat:
@bgrenon thank you for your comments 🙏 
All your requests resolved, merge conflicts as well

bgrenon:
@saimonkat just updated the sidebar.yaml with the latest structure we'd like to use. Side note: _really_ like how the logic works in the sidebar; it's super-flexible for us to modify things to best fit the content.

A few further tweaks. 
NOTE: these are non-blocking. It's looking good, and we can refine it later. But adding these here in case they are simple fixes:

1. Duplicated pages break the sidebar. Example: 
https://neon-next-git-docs-style-update-neondatabase.vercel.app/docs/extensions/pgvector
This page is included in two secondary navs, which actually is something we'd like to do. The Framework guides for example: we'd like them both [here](https://neon-next-git-docs-style-update-neondatabase.vercel.app/docs/get-started-with-neon/signing-up) and under "Develop" in the main nav.
If this is a longer fix, we can work around this for now.

2. Not sure if this works, but consider moving the “Home” to the top of the page. And instead of using breadcrumb-links to the previous nav in the sidebar, replace that permanently with Home. And then rely on a more obviously styled breadcrumb in the body to let users go back one level.
Example:
image

3. I’m also not sure the “<— back to site” link at the bottom of the sidebar works. It’s not clear the site is neon.tech/home or the docs site. Later, I wonder if there is way to get a link to “neon.tech/home” in the header of the page, maybe a hamburger with other key links?

4. An icon for the "Neon roadmap" which is now in the main nav.

saimonkat:
Hi @bgrenon thank you for your updates for the sidebar structure and your comments 🙏 

1. Duplicated pages break the sidebar.
We didn't assume we'll have duplicate pages. So I'll take a look if it will be a quick fix. 
Now I have this [getActiveItems](https://github.com/neondatabase/website/blob/61a670e04803d4b26dc0eab0cda7afaf17d7bf92/src/components/pages/doc/sidebar/sidebar.jsx#L18) function to define initial sidebar opened state, will try to modify it. 
By the way, which page do you want to open by the duplicated slug? Is the first found page works?

2-3. As I remember, all back links ("Home", breadcrumb and “<— back to site”) were requested from business and we also worked on them for the better UX. We'll discuss there points with our team.

4. We'll draw and update an icon for the "Neon roadmap" which is now in the main nav, no problems 🙌

saimonkat:
Hi @bgrenon 
Here are updates for the last 4 tasks we discussed:

1. Duplicated pages. I researched on Friday and tried to implement this fix. However it turned out not to be a quick fix, because it's changing our active menus calculation radically. I'll need a couple of hours for this task and it would be better to implement it after this PR release. Therefore, if this is not so critical for you, we suggest not duplicating pages in this release.

2-3. We'll discuss these points with our team today. 

4. Icon for roadmap updated. Hope you'll like it
    image

danieltprice (approved):

PR #1925, opened 2024-07-11
Updating max compute limit from 8 to 10 CU/vCPU

bgrenon:
Docs updated to capture new max compute limit of 10 CU/vCPU. Which means 2 additional tiers of CU only available on Scale plan and above:

- 9 CU: 9 vCPU, 36 GB RAM
- 10 CU: 10 vCPU, 40 GB RAM

Note: need to check the 9 and 10 CU numbers for [maintenance_work_mem](https://neon-next-git-bgrenon-update-max-compute-limit-neondatabase.vercel.app/docs/extensions/pgvector#maintenance_work_memto) defaults once its in production


danieltprice (approved):

danieltprice:
9 CU is not an option in the UI, but can it be set via the API, and do we document it?

maintenance_work_mem
9 CU = ?
10 CU = 687590kB (671.86MB)

max_connections
9CU = ?
10 CU = 4000

danieltprice (approved):

PR #1952, opened 2024-07-20
changelog-07-26-2024

danieltprice:
preview:
https://neon-next-git-07-26-2024-changelog-neondatabase.vercel.app/docs/changelog


danieltprice:
- CLI updates
- Drizzle studio updates

davidgomes (approved):
The `create-app` related changes are great! (especially the screenshots)

content/changelog/2024-07-26.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Clearer language around compute types
+
+Words matter. We've changed our naming convention around compute types: from `RW compute` and `RO Replica` to a cleaner, more straightforward `Primary compute` and `Read replica`.
+
+![new naming for compute types](/docs/relnotes/compute_types.png)
+
+With Neon's unique architecture, where we separate storage from compute for copy-on-write branching, you can choose the size and features for the compute that powers your branch's database independently from your data.
+
+- **Primary compute** &#8212; You don't have to add a compute to your branch right away, but once you decide to connect to your database, you need a Primary compute. Think of your primary compute as the engine that runs your branch. By adding a primary compute, you are in essence turning your branch "on". By default, your primary compute is read-write, though you can modify database permissions using traditional Postgres roles.
```
danieltprice, line 13: Maybe rework this description a bit. We're also missing a glossary definition for this. 
A few points to consider:
- "You don't have to add a compute to your branch right away": From the Neon Console, you always get a primary compute when you create a branch. Most users won't have the experience of creating a branch without a compute. You can only do this via the API. You can delete a compute via the console, but you'll always start with a branch that has a primary compute. You don't have to add one to connect - it'll already be there for most users.
- "once you decide to connect to your database, you need a Primary compute" - you can also connect to a branch via a read-replica compute so this is not entirely accurate.
- "By default, your primary compute is read-write" - this will always be the case. A primary compute can only be RW and you cannot "change the default" for this compute.
- "Think of your primary compute as the engine that runs your branch" - this is true, but both primary and read replica computes run postgres and can act as an "engine". 






-

content/changelog/2024-07-26.md
```diff
@@ -0,0 +1,31 @@
...
+- **Read replicas** &#8212; When you're ready to scale your application, you can offload read-only traffic to one or more read replicas. In Neon, read replicas work differently than traditional systems because the data is not physically replicated. Primary computes and read replicas work from the same data &#8212; a single, cost-effective source of truth.
```
danieltprice, line 14: I think we could drop "a single, cost-effective source of truth". I'm not quite sure how to think about the "source of truth" part.

content/changelog/2024-07-26.md
danieltprice: Some points about primary computes & read replicas:

Primary compute

A read-write compute runs Postgres.
Each branch in a Neon project can have a single primary compute. 
A Neon project is created with a default branch with an associated primary compute.
When you add branches via the Neon Console, each branch is created with a primary compute.

Read replica

A read-only compute runs Postgres. 
Read replicas let you scale your application by offloading read-only traffic from your primary read-write compute.
Each branch in a Neon project can have one or more read replicas.
Read replicas are designed to perform read operations on the same data as your primary read-write compute. Read replicas are asynchronous, which means they are eventually consistent. 
For more, see [Read replicas](/docs/introduction/read-replicas).

content/changelog/2024-07-26.md
bgrenon: Revised.

bgrenon (approved):

bgrenon (approved):

PR #2064, opened 2024-08-22
Move Prisma Pulse guide to the correct section

sneub:
Hey @danieltprice 👋  just noticed that the Prisma Pulse guide was moved into the wrong section (I think) -- I believe it would fit better in the _Replicate data **from**_ section rather than the _**to**_ section.


danieltprice:
@sneub Thank you for reporting! We have you in the right section via another PR:
https://neon.tech/docs/guides/integrations#replicate-data-from-neon

Closing this PR.

PR #2137, opened 2024-09-06
Update hero.jsx

carlotas19:
- Fix a mistake in the hero (number of branches should be off for all plans, now it shows in 2 / 4) 
- Fix a mistake in the detailed table (IP Allow is not available in Scale no more)


YanicNeon (approved):
Looks good to me, thank you for the fast reaction !

PR #2298, opened 2024-10-10
ci: report workflow stats

fedordikarev:


bayandin (commented):
I don't think we're interested in collecting stats for this repo at the moment

PR #2319, opened 2024-10-16
Private Access docs

danieltprice:
Preview:
https://neon-next-git-dprice-private-link-neondatabase.vercel.app/docs/guides/neon-private-access

Notes:
- The steps have not been validated. The document is purely based on the provided Notion doc.
- There is a request with Pixel Point to provide a Neon-style version of the diagram


content/docs/guides/neon-private-access.md
```diff
@@ -0,0 +1,97 @@
...
+
+2. **Provide the endpoint ID to Neon**
+
+   Note the ID of the created endpoint and provide it to Neon. Neon will authorize this endpoint to access the Neon Private Access service and notify when the authorization is complete.
+
+3. **Enable Private DNS**
+
+   After Neon authorizes your endpoint (please wait for confirmation from Neon), you need to enable private DNS lookup for the endpoint.
+
+   1. In AWS, select the VCP endpoint you created.
```
stradig, line 57: This must be VPC instead of VCP - my bad.

content/docs/guides/neon-private-access.md
```diff
@@ -0,0 +1,97 @@
...
+
+<PrivatePreview />
+
+The **Neon Private Access** feature lets you securely connect to your Neon databases using [AWS PrivateLink](https://docs.aws.amazon.com/vpc/latest/privatelink/concepts.html), avoiding the open internet for improved security.
+
+## Overview
+
+In a standard setup, the client application connects to a Neon database over the open internet via the Neon proxy.
+
+With **Neon Private Access**, you can connect to your database via AWS PrivateLink instead of over the open internet. In this configuration, the private access client connects to a private instance of the Neon proxy through an [AWS endpoint service](https://docs.aws.amazon.com/vpc/latest/privatelink/configure-endpoint-service.html) provided by Neon. This endpoint service is accessible only within the same AWS region as your client and is restricted to Neon-authorized customers. With **Neon Private Access**, all traffic between the private access client application and the Neon database remains within AWS’s private network, rather than traversing the public internet.
```
stradig, line 16: "connects to a private instance": This could convey the message that it is per customer, which is not the case. We do not want to advertise that this is not the case, but we also would not want to mislead people. I would use "dedicated instance", maybe?

stepashka: perhaps we could say something like
'connects to an instance of Neon proxy that is not accessible from the public internet'?

stradig: Sounds good, yes!

content/docs/guides/neon-private-access.md
danieltprice: fixed.

PR #2343, opened 2024-10-21
Update database-access.md

scottcate:
Bug Fix / Typo. `CREATER ROLE` should be `CREATE USER` in two locations.

### DIFF
--- a/content/docs/manage/database-access.md
+++ b/content/docs/manage/database-access.md
@@ -159,7 +159,7 @@ To create a read-write role:
 4. Create a database user. The password requirements mentioned above apply here as well.
 
    ```sql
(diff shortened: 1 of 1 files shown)


ruf-io:
@scottcate Thanks for submitting this - I was about to hit approve and then realized we wrote about this a while ago...

https://neon.tech/blog/postgres-roles

Postgres roles can be counterintuitive:

> Any role in Postgres can be granted the ability to log in. When a role has this capability, it effectively becomes what we traditionally think of as a user. This is typically done with: `CREATE ROLE username LOGIN`

image

So I think both your new suggested code and the old code work, but the canonical way is with `ROLE`

PR #2355, opened 2024-10-23
docs: fixed guide formatting issues caused by prettier

sam-harri:
Weird issue where when the Prettier bot formats the md guides, it undoes the <CodeTabs> block triple backticks, and replaces them with oddly formatted single backticks. If I indent the CodeTabe block once more than it should be, Prettier leaves it as is, and it still renders the same on the website


ruf-io (approved):
LGTM - Let's see if prettierbot lets this one through :)

PR #2388, opened 2024-10-28
docs: make the syntax highllighting more appropriate for the content

IAmJSD:


content/postgresql/postgresql-administration/create-superuser-postgresql.md
```diff
@@ -50,13 +50,13 @@ PASSWORD 'moreSecurePass';
 
 Third, verify the user creation:
 
-```sql
+```text
```
rishi-raj-jain, line 53: why text? isn't it s a SQL query?

ruf-io: Seems like the SQL syntax accounts for backslash commands but I agree with @IAmJSD this is psql syntax not part of SQL. Let's keep it text

ruf-io (approved):
LGTM

PR #2391, opened 2024-10-29
Fix table, trim header links in Neon Authorize docs

bgrenon:


danieltprice:
lbraces are not rendering properly:

Connection times may be slow during this process.

PR #2400, opened 2024-10-30
docs: add error list and links to top of page

bgrenon:
Preference is to use error messages as heading equivalent on this page, which breaks the rightnav.

This is a workaround to give a list of errors a the top of the page; each error links to an anchor in the relevant section.


danieltprice (approved):
LGTM!

PR #2412, opened 2024-10-31
Adding Stack Auth guide, plus general edits to other authorize content

bgrenon:
[Stack Auth quickstart](https://neon-next-git-bgrenon-stackauth-neonauthorize-neondatabase.vercel.app/docs/guides/neon-authorize-stack-auth)
[Edits to Clerk quickstart](https://neon-next-git-bgrenon-stackauth-neonauthorize-neondatabase.vercel.app/docs/guides/neon-authorize-clerk)


btholt:
Left comments directly on Vercel

btholt (changes requested):
Left comments directly on Vercel

content/docs/guides/neon-authorize-clerk.md
```diff
@@ -39,89 +38,98 @@ In this first set of steps, we’ll integrate Clerk as an authorization provider
...
 GRANT SELECT, UPDATE, INSERT, DELETE ON ALL TABLES IN SCHEMA public TO anonymous;
 ```
 
-Later, you can define RLS policies that will restrict what your application's users can do with these roles.
+- **Authenticated role**: This role is intended for users who are logged in. Your application should send the authorization token when connecting using this role.
+- **Anonymous role**: This role is intended for users who are not logged in. It should allow limited access, such as reading public content (e.g., blog posts) without authentication.
 
 ### 5. Install the Neon Serverless Driver
 
 Neon’s Serverless Driver manages the connection between your application and the Neon Postgres database. It supports HTTP, WebSockets, and TCP connections. For Neon Authorize, the connection must be established over HTTP. For HTTP, we recommend using this driver.
```
danieltprice, line 89: I wonder if we should avoid saying "recommend" unless there are other options. Are there other drivers that could be used?

danieltprice (approved):

content/docs/guides/neon-authorize-clerk.md
bgrenon: I can reword to make it clearer. In theory "the HTTP API can be used from any language and Authorize can be used with the HTTP API from any language" but we want to recommend using our driver.

`Neon’s Serverless Driver manages the connection between your application and the Neon Postgres database. For Neon Authorize, HTTP is requrired. While it is technically possible to access the HTTP API without using our driver, we recommend using the driver for best performance. The driver also supports WebSockets and TCP connections, so make sure you use the HTTP method when working with Neon Authorize.`

PR #2451, opened 2024-11-11
Add sparkline img api for emails

ruf-io:
Adds a image generator API for creating sparklines like this: https://neon.tech/api/sparkline?series=10,20,10,0,15,25&w=120&h=24

should generate something like this 
image


saimonkat:
Hi @ruf-io do you consider this PR is still relevant?

PR #2519, opened 2024-11-22
docs: add missing /api prefix to stack-auth jwks example route

amcgee:
The example route is missing the `/api` prefix, causing a 404 Not Found error (after substituting a valid project ID)


bgrenon (approved):
Good catch. Thank you for contributing!

PR #2545, opened 2024-11-26
docs: fix grammar in BETWEEN example conclusion

frontendninja10:
The example description for the LENGTH() function incorrectly used "gets" instead of "to get". This commit improves the grammar for enhanced clarity.

Before:
"In this example, we use the LENGTH() function gets the number of characters of an input string."

After:
"In this example, we use the LENGTH() function to get the number of characters of an input string."


danieltprice (approved):

danieltprice:
Thank you @frontendninja10!

PR #2546, opened 2024-11-26
docs: fix grammar in explanation of the WHERE clause

frontendninja10:
The sentence "cause the condition evaluates to true" has been corrected to "cause the condition to evaluate to true".

Before:
"In other words, the query will include only rows that cause the condition evaluates to true in the result set."

After:
"In other words, the query will include only rows that cause the condition to evaluate to true in the result set."


danieltprice (approved):

danieltprice:
Thank you @frontendninja10 !

PR #2550, opened 2024-11-27
add NAT gateway documentation for new Azure regions

fcdm:
Adds public IPs for NAT gateways in prod Azure `gwc` and `westus3`.

### DIFF
--- a/content/docs/introduction/regions.md
+++ b/content/docs/introduction/regions.md
@@ -74,9 +74,9 @@ If you are unsure of your project's region, you can find this information in the
 
 ### Azure NAT Gateway IP Addresses
 
-| Region                                     | NAT Gateway IP Addresses                       |
-| :----------------------------------------- | :--------------------------------------------- |
-| Azure East US 2 (Virginia) — azure-eastus2 (**Beta**) | 48.211.218.176, 48.211.218.194, 48.211.218.200 |
+| Region                                                      | NAT Gateway IP Addresses                       |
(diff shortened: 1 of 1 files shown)


content/docs/introduction/regions.md
```diff
@@ -21,6 +21,8 @@ Neon supports project deployment in several regions. We recommended that you sel
 ## Azure regions
 
 - Azure East US 2 region (Virginia) &mdash; `azure-eastus2` (**Beta**)
+- Azure Germany West Central region &mdash; `azure-gwc` (**Beta**)
```
fcdm, line 24: Beta labels might be going away soon but I'm not sure about the timeline.

content/docs/introduction/regions.md
```diff
@@ -21,6 +21,8 @@ Neon supports project deployment in several regions. We recommended that you sel
...
+- Azure West US 3 region (Arizona) &mdash; `azure-westus3` (**Beta**)
```
fcdm, line 25: I've added the state name but we're not actually showing them in console, e.g. it's just `East US 2`, so maybe we can remove them?

danieltprice (approved):

content/docs/introduction/regions.md
danieltprice: We reference this page from the Neon API docs, where you have to specify a region_id value like aws-us-east-1, although, I have not tried the Create project API with an Azure region ID. I should test that. I assume it works.

PR #2552, opened 2024-11-27
Fix data types prev and next urls

sdarnadeem:
Some previous and next urls were not pointing to the correct pages as per the index. This pull requests fixes all these pages which have incorrect previous and next linking in data types category.


philip:
Thank you for fixing these @sdarnadeem !

PR #2585, opened 2024-12-03
chore: use static redirects instead of directly in SSR 

amitdahan:


next.config.js
```diff
@@ -245,6 +245,40 @@ const defaultConfig = {
...
+        permanent: false,
+      },
+      {
+        source: '/',
+        has: [{ type: 'cookie', key: 'neon_login_indicator' }],
+        missing: [
+          { type: 'header', key: 'referer' },
+          { type: 'header', key: 'referrer' },
+        ],
+        destination: 'https://console.neon.tech',
```
rishi-raj-jain, line 273: afaik it would be the same as running those in the middleware.js file, since the path matching and redirects config are invoked before cache is served.

saimonkat:
Hi @amitdahan @rishi-raj-jain we've refactored this PR with moving redirects to middleware: https://github.com/neondatabase/website/pull/2597

Key points:
- remove checking from SSR, but still keep it dynamic and asynchronous
- easily incorporate environment-specific logic using `process.env.NODE_ENV`
- implement error handling and logging within middleware
- implement complex conditions, such as checking multiple headers or cookies, and perform different actions based on their values

amitdahan:
Closing since we already merged #2597

PR #2597, opened 2024-12-04
refactor: move redirects from homepage to middleware

saimonkat:
This PR refactors homepage redirects, moving them to middleware

reference: https://github.com/neondatabase/website/pull/2585

basic concepts:
- Root Path (`/`): In production, if the user is logged in, they are redirected to /home if the referer matches the current branch or default site URL; otherwise, they are redirected to the console.
- Home Path (`/home`): Users not logged in are redirected to the root path (/).

[Preview](https://neon-next-git-move-redirects-to-middleware-neondatabase.vercel.app/)

### DIFF
--- a/src/app/(home-page)/home/page.jsx
+++ b/src/app/(home-page)/home/page.jsx
@@ -1,6 +1,3 @@
-import { redirect } from 'next/navigation';
-
-import { checkCookie } from 'app/actions';
 import AiIndex from 'components/pages/home/ai-index';
 import Bento from 'components/pages/home/bento';
 import Hero from 'components/pages/home/hero';
@@ -19,27 +16,20 @@ export const metadata = getMetadata({
   robotsNoindex: 'noindex',
 });
 
-const HomePage = async () => {
-  const is_logged_in = await checkCookie('neon_login_indicator');
-  if (!is_logged_in) {
-    return redirect('/');
-  }
-
-  return (
-    <>
-      <Hero />
-      <Logos />
-      <InstantProvisioning />
-      <Lightning />
-      <Bento />
-      <AiIndex />
-      <Multitenancy />
-      <Industry />
-      <Trusted />
-      <Cta />
-    </>
-  );
-};
+const HomePage = () => (
+  <>
+    <Hero />
+    <Logos />
+    <InstantProvisioning />
+    <Lightning />
+    <Bento />
+    <AiIndex />
+    <Multitenancy />
+    <Industry />
+    <Trusted />
+    <Cta />
+  </>
+);
 
 export default HomePage;
 
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -101,3 +101,7 @@ export async function middleware(req) {
     return NextResponse.redirect(new URL(SITE_URL));
   }
 }
+
+export const config = {
+  matcher: ['/', '/home', '/generate-ticket/:path*', '/tickets/:path*'],
+};
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -11,7 +11,6 @@ const SITE_URL =
     ? `https://${process.env.VERCEL_BRANCH_URL}`
     : process.env.NEXT_PUBLIC_DEFAULT_SITE_URL;
 
-const homeProtectedRoutes = ['/', '/home'];
 const ticketsProtectedRoutes = ['/generate-ticket', '/tickets'];
 
 const extractHandleFromPath = (pathname) => pathname.split('/').slice(-2)[0];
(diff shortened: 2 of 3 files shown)


saimonkat:
SpeedVitals comparison for America

[neon.tech](https://speedvitals.com/ttfb-test?url=https://neon.tech) - medium 400
image

[This PR Preview](https://app.speedvitals.com/ttfb-test?url=https://neon-next-git-move-redirects-to-middleware-neondatabase.vercel.app%2F) - medium 200
image

rishi-raj-jain (changes requested):
as mentioned in Slack:

- the middleware runs on every route, not just / and /home like the previous redirects were
- first - IIUC we'll always check the indicator, not only on the /home and / routes?
- second (maybe due to the above) - we're missing the /home -> /  redirect (if the cookie indicator is missing)

rishi-raj-jain:
@saimonkat What do you think about exporting a matcher in the middleware to only run it in on the homepage routes (/ and /home)?

saimonkat:
@rishi-raj-jain I guess we have it already [here](https://github.com/neondatabase/website/pull/2597/files#diff-9a117d52a8a2cb3e5f498575bdfdea4dc32418180c09029454fbaf7f95ebe5a1R35)

```
const homeProtectedRoutes = ['/', '/home'];
...
if (homeProtectedRoutes.includes(pathname)) {
...
```

or do you mean another approach?

rishi-raj-jain:
But that invokes middleware for each path, correct?

saimonkat:
@rishi-raj-jain oh I see…
but we have in this file also tickets middlewares: 
`const ticketsProtectedRoutes = ['/generate-ticket', '/tickets'];`

do you want to split them to 2 separate files?

rishi-raj-jain:
> @rishi-raj-jain oh I see… but we have in this file also tickets middlewares: `const ticketsProtectedRoutes = ['/generate-ticket', '/tickets'];`
> 
> do you want to split them to 2 separate files?

Nah, let's just run that logic on those 4 paths only, saves us from running that logic in all the paths as happens currently.

saimonkat:
@rishi-raj-jain updated ✅ 
please check

rishi-raj-jain:
I think now we can skip the checks in https://github.com/neondatabase/website/blob/98e160f5c9e801072245d4f673f2a0e3dc46d8f2/src/middleware.js#L25-L32?

src/app/(home-page)/home/page.jsx
```diff
@@ -19,27 +16,20 @@ export const metadata = getMetadata({
   robotsNoindex: 'noindex',
```
rishi-raj-jain, line 16: why do we make this noindex?

rishi-raj-jain (approved):

src/middleware.js
```diff
@@ -16,17 +20,30 @@ const generateEditPageURL = (handle) => `${SITE_URL}/tickets/${handle}/edit`;
...
+          if (
+            referer.includes(process.env.VERCEL_BRANCH_URL) ||
+            referer.includes(process.env.NEXT_PUBLIC_DEFAULT_SITE_URL)
+          ) {
+            return NextResponse.redirect(new URL(`${SITE_URL}/home`));
+          }
+        } catch (error) {
+          console.error('Error getting referer:', error);
+        }
+        return NextResponse.redirect(new URL(LINKS.console));
```
amitdahan, line 38: AFAIK `redirect` accepts a `string` as well as a `URL`, and since `LINKS.console` is a full URL string, this should simply be enough:

```suggestion
        return NextResponse.redirect(LINKS.console);
```

src/middleware.js
```diff
@@ -16,17 +20,30 @@ const generateEditPageURL = (handle) => `${SITE_URL}/tickets/${handle}/edit`;
...
+    try {
+      const isLoggedIn = await checkCookie('neon_login_indicator');
+      if (pathname === '/' && isLoggedIn) {
+        try {
+          const referer = await getReferer();
+          if (
+            referer.includes(process.env.VERCEL_BRANCH_URL) ||
+            referer.includes(process.env.NEXT_PUBLIC_DEFAULT_SITE_URL)
+          ) {
+            return NextResponse.redirect(new URL(`${SITE_URL}/home`));
```
amitdahan, line 33: Seems like this would work, but either:
- `new URL` is redundant here
- or let's use `new URL`'s 2nd parameter

so:
```suggestion
            return NextResponse.redirect(`${SITE_URL}/home`);
```

or:
```suggestion
            return NextResponse.redirect(new URL('/home', SITE_URL));
```

---

But actually [docs](https://nextjs.org/docs/app/building-your-application/routing/middleware#example) mention just using `req.url` as the base:

```suggestion
            return NextResponse.redirect(new URL('/home', req.url));
```

amitdahan (approved):
LGTM, left 2 minor comments around `NextResponse.redirect` calls.

PR #2619, opened 2024-12-09
chore: add sign up events

bacher:
Relates to https://linear.app/neondb/issue/GRW-290/experiment-build

Add 2 required for the experiment events:

- `header_sign_up_clicked`
- `home_hero_get_started_clicked`

### DIFF
--- a/src/components/pages/home/hero/hero.jsx
+++ b/src/components/pages/home/hero/hero.jsx
@@ -131,6 +131,7 @@ const Hero = () => {
             to={LINKS.signup}
             target="_blank"
             tag_name="Hero"
+            analyticsEvent="home_hero_get_started_clicked"
           >
             Get Started
           </Button>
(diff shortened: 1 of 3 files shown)


ruf-io:
We already have a way of adding tracking events, and these events should already be tracked. This PR should be unnecessary.

These should be tracked via "Button Clicked" events that are added at the button level.

bacher:
> We already have a way of adding tracking events, and these events should already be tracked. This PR should be unnecessary.
> 
> These should be tracked via "Button Clicked" events that are added at the button level.

x-post from slack:
Yeah, I saw those events, but the problem with them is that they are not visible in Posthog, since we didn't set up Segment <-> Posthog integration so far. It works only for Console right now.
Segment <-> Posthog integration is way more effort to do. I guess we eventually will get there, but it's too overcomplicated to do it for current experiment.
So, I will remove all those events from the codebase once the experiment is considered finished.

duskpoet (approved):

PR #2678, opened 2024-12-18
Update postgresql-nth_value-function.md

zvijayakumar:
it is mentioned as the most expensive. It is wrong. it should be the second most expensive product

Please check the query and description


danieltprice (approved):

danieltprice:
Thanks for the fix @zvijayakumar !

PR #2718, opened 2025-01-01
Update correct next link in postgresql-having.md

da-vaibhav:
https://neon.tech/postgresql/postgresql-tutorial/postgresql-having has next link as https://neon.tech/postgresql/postgresql-tutorial/postgresql-union but sidebar indicates that next link should be postgresql-grouping-sets


danieltprice (approved):
Thank you @da-vaibhav !

PR #2722, opened 2025-01-02
Update postgresql-drop-procedure.md

frontendweb-pk:
Replace static values to variables


danieltprice (approved):
Hi @frontendweb-pk 
Thanks for your suggestion!

PR #2723, opened 2025-01-02
Fix typo in postgresql-rename-table.md

kofta999:


danieltprice (approved):
Thank you for the fix @kofta999 !

kofta999:
Np! and thank you all for the awesome postgres tutorial, really liking it so far.

PR #2740, opened 2025-01-07
Update pl-pgsql-assert.md

frontendweb-pk:
Replace > to <


danieltprice (approved):
Thank you @frontendweb-pk

PR #2769, opened 2025-01-10
chore: force disable "website_growth_compute_price_rising" exp

bacher:
Relates to [linear.app/neondb/issue/GRW-290/experiment-build](https://linear.app/neondb/issue/GRW-290/experiment-build)

I want to disable this experiment in code and run experiment again to check posthog's results in this case, where we 100% sure that there are no any actual changes.

The previous results of experiment is very suspicious for me https://us.posthog.com/project/49356/experiments/68916

Later I will remove this exp from codebase completely.

### DIFF
--- a/src/components/pages/pricing/hero/hero.jsx
+++ b/src/components/pages/pricing/hero/hero.jsx
@@ -34,7 +34,7 @@ const scaleCardBorderVariants = {
 const Hero = () => {
   const posthog = usePostHog();
   const isComputePriceRaised =
-    useFeatureFlagVariantKey('website_growth_compute_price_rising') === 'show_0_24';
+    useFeatureFlagVariantKey('website_growth_compute_price_rising') === 'show_0_24' && false;
 
   const plans = useMemo(() => {
     if (isComputePriceRaised) {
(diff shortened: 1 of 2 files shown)


duskpoet:
lets this time do a manual experiment? With segment events and charts in superset/hex

bacher:
> Why not just disable the FF?

I want to disable this experiment in code and run experiment again to check posthog's results in this case, where we 100% sure that there are no any actual changes.

bacher:
> lets this time do a manual experiment? With segment events and charts in superset/hex

I don't want to repeat the experiment, since we don't want to show $0.24 to users anymore.

This PR is preparing for me running test experiment, I just want to look at posthog results for such void experiment. So ideally if UI is not changed (since feature flags evaluated but not used on frontend side), then results should show that both variants show the same metrics, but it seems that Posthog can show difference anyway. I want to check it.

duskpoet (approved):
I don't think you will get any meaningful results from such experiment.
Posthog will still vary flag values and measure conversion. And if number of events is statistically insignificant the results will be random anyway

PR #2829, opened 2025-01-26
Fix: postgresql-any.md SELECT query output.

FadyMorris:
Fix the output of the query:
```sql
SELECT * FROM employees;
```


danieltprice (approved):

danieltprice:
Thank you @FadyMorris

PR #2830, opened 2025-01-26
Update postgresql-index-on-expression.md

desouky22:
There is an index on the last_name not the first_name in the customer table


danieltprice (approved):

danieltprice:
Thank you @desouky22

PR #2855, opened 2025-01-30
Update postgresql-limit.md

manu-mathani:
I included the 'DESC' clause in the first explanation statement of the last example.


danieltprice (approved):
Thank you @manu-mathani!

PR #2889, opened 2025-02-05
Rename Neon Identity to Neon Auth + best practice around foreign keys

bgrenon:
https://neon-next-git-bgrenon-rename-identity-neondatabase.vercel.app/docs/guides/neon-auth

### DIFF
--- a/content/docs/guides/neon-auth.md
+++ b/content/docs/guides/neon-auth.md
@@ -140,7 +140,7 @@ Here's an example of Neon Auth with Stack Auth. As we add more providers and fea
 
 ### Foreign keys and the users_sync table
 
-Since the `neon_auth.users_sync` table is updated asynchronously, there may be a brief delay (usually less than 1 second) before a user's data appears in the table. This means foreign key constraints could temporarily fail for new users. Consider this possible delay when deciding whether to use foreign keys in your schema.
+Since the `neon_auth.users_sync` table is updated asynchronously, there may be a brief delay (usually less than 1 second) before a user's data appears in the table. Consider this possible delay when deciding whether to use foreign keys in your schema.
 
 If you do choose to use foreign keys, make sure to specify an `ON DELETE` behavior that matches your needs: for example, `CASCADE` for personal data like todos or user preferences, and `SET NULL` for content like blog posts or comments that should persist after user deletion.
 
--- a/content/docs/guides/neon-auth.md
+++ b/content/docs/guides/neon-auth.md
@@ -167,7 +167,7 @@ CREATE TABLE posts (
 
 When querying data that relates to users:
 
-- Use LEFT JOINs instead of INNER JOINs with the `users_sync` table in case of any sync delays. This ensures that all records from the main table (.e.g. posts) are returned even if there's no matching user in the `users_sync` table yet.
+- Use LEFT JOINs instead of INNER JOINs with the `users_sync` table in case of any sync delays. This ensures that all records from the main table (e.g., posts) are returned even if there's no matching user in the `users_sync` table yet.
 - Filter out deleted users since the table uses soft deletes (users are marked with a `deleted_at` timestamp when deleted).
 
 Here's an example of how to handle both in your queries:
--- a/content/docs/guides/neon-auth.md
+++ b/content/docs/guides/neon-auth.md
@@ -3,6 +3,8 @@ title: About Neon Auth
 subtitle: Automatically sync user profiles from your auth provider directly to your
   database
 enableTableOfContents: true
+redirectFrom:
+  - /docs/guides/neon-identity
 tag: beta
 updatedOn: '2025-01-31T21:21:32.228Z'
 ---
@@ -25,7 +27,7 @@ Neon Auth solves this by integrating your auth provider with your Postgres datab
 
 ## How Neon Auth works
 
-When you set up Neon Auth, we create a `neon_identity` schema in your database. As users authenticate and manage their profiles in your auth provider, their data is automatically synchronized to your database.
+When you set up Neon Auth, we create a `neon_auth` schema in your database. As users authenticate and manage their profiles in your auth provider, their data is automatically synchronized to your database.
 
 Here is the basic flow:
 
@@ -41,10 +43,10 @@ Here is the basic flow:
 
 3. **The data is immediately available in your database**
 
-   The synchronized data is available in the `neon_identity.users_sync` table shortly after the auth provider processes changes. Here's an example query to inspect the synchronized data:
+   The synchronized data is available in the `neon_auth.users_sync` table shortly after the auth provider processes changes. Here's an example query to inspect the synchronized data:
 
    ```sql
-   SELECT * FROM neon_identity.users_sync;
+   SELECT * FROM neon_auth.users_sync;
    ```
 
    | id          | name       | email             | created_at    | raw_json                      |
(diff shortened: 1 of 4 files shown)


danieltprice (approved):

content/docs/guides/neon-identity.md
```diff
@@ -9,23 +9,23 @@ updatedOn: '2025-01-31T21:21:32.228Z'
...
 
 - Provision auth provider projects and manage your users directly from the Neon Console
 - Automated synchronization of user profiles between auth provider and your database
 - Easy database relationships, since your user profiles are available as any other table
 
-## How Neon Identity works
+## How Neon Auth works
 
-When you set up Neon Identity, we create a `neon_identity` schema in your database. As users authenticate and manage their profiles in your auth provider, their data is automatically synchronized to your database.
+When you set up Neon Auth, we create a `neon_identity` schema in your database. As users authenticate and manage their profiles in your auth provider, their data is automatically synchronized to your database.
```
pffigueiredo, line 28: `neon_auth`, we also changed the `schema` name

content/docs/guides/neon-identity.md
```diff
@@ -69,18 +69,18 @@ Updates to user profiles in the auth provider are automatically synchronized.
 Do not try to change the `neon_identity.users_sync` table name. It's needed for the synchronization process to work correctly.
```
pffigueiredo, line 69: `neon_auth.users_sync`

content/docs/guides/neon-identity.md
```diff
@@ -110,9 +110,9 @@ jobs.on('user.updated', async (event) => {
 });
 ```
 
-### After Neon Identity
+### After Neon Auth
 
-With Neon Identity, much of this complexity is eliminated. Since user data is automatically synced to `neon_identity.users_sync`, you can just create the todo:
+With Neon Auth, much of this complexity is eliminated. Since user data is automatically synced to `neon_identity.users_sync`, you can just create the todo:
```
pffigueiredo, line 115: `neon_auth.users_sync`

content/docs/guides/neon-identity.md
```diff
@@ -124,12 +124,41 @@ async function createTodo(userId: string, task: string) {
 }
 ```
 
-## Getting Started
+## Getting started
 
 1. From the Neon Console, navigate to the **Identity** tab. Choose your provider and click **Connect**. Currently, only Stack Auth is available for Early Access users.
```
pffigueiredo, line 129: THis isn't exactly like this right now, but might as well just change it to `Auth` tab

content/docs/guides/neon-identity.md
```diff
@@ -124,12 +124,41 @@ async function createTodo(userId: string, task: string) {
...
-Here's an example of Neon Identity with Stack Auth. As we add more providers and features, this interface will continue to evolve.
+Here's an example of Neon Auth with Stack Auth. As we add more providers and features, this interface will continue to evolve.
 
 ![identity with stackauth deployed](/docs/guides/identity_stackauth.png)
+
+## Best practices
+
+### Foreign keys and the users_sync table
+
+Since the `neon_identity.users_sync` table is updated asynchronously, there may be a brief delay (usually at most a few seconds) before a user's data appears in the `users_sync` table. You may need to implement retry logic or error handling in your application for database operations that reference new users.
```
pffigueiredo, line 141: neon_auth.users_sync

content/docs/guides/neon-identity.md
```diff
@@ -124,12 +124,41 @@ async function createTodo(userId: string, task: string) {
...
+## Getting started
 
 1. From the Neon Console, navigate to the **Identity** tab. Choose your provider and click **Connect**. Currently, only Stack Auth is available for Early Access users.
 1. You'll be asked to authenticate and select the project you want to integrate with.
 1. Once connected, you'll see the integration view. This shows your synced users, connection status, and quick links to your provider's documentation and console to help configure your application (e.g. SSO or API keys)
 
-Here's an example of Neon Identity with Stack Auth. As we add more providers and features, this interface will continue to evolve.
+Here's an example of Neon Auth with Stack Auth. As we add more providers and features, this interface will continue to evolve.
 
 ![identity with stackauth deployed](/docs/guides/identity_stackauth.png)
```
pffigueiredo, line 135: change the `alt` text to `auth with stackauth deployed`

content/docs/guides/neon-identity.md
pffigueiredo, line 141: > here may be a brief delay (usually at most a few seconds)

From reading this, I get the impression that it might take a few seconds, which indeed might happen, but usually it will almost be instant, so, under 1 second kind of thin.

> before a user's data appears in the

"before a user's data is synced in the ..."

> You may need to implement retry logic or error handling in your application for database operations that reference new users.

More than this, I think it's important to notice that consumers need to make sure that their application won't break if the user hasn't been synced yet.

content/docs/guides/neon-identity.md
```diff
@@ -124,12 +124,41 @@ async function createTodo(userId: string, task: string) {
...
+);
+
+-- For content that should persist after user deletion (e.g., blog posts)
+CREATE TABLE posts (
+    id SERIAL PRIMARY KEY,
+    title TEXT NOT NULL,
+    content TEXT NOT NULL,
+    author_id UUID REFERENCES neon_identity.users_sync(id) ON DELETE SET NULL,
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+);
```
pffigueiredo, line 161: Some other best practices could be:

- use left joins when relating with the `users_sync` table so that you only get data whose user is already fully synced
- since we use soft deletes, don't forget to filter your users data by `deleted_at`

content/docs/sidebar.yaml
```diff
@@ -489,7 +489,7 @@
           slug: guides/ruby-on-rails
         - title: SQLAlchemy
           slug: guides/sqlalchemy
-    - title: Neon Identity
+    - title: Neon Auth
       slug: guides/neon-identity
```
pffigueiredo, line 493: this `slug` won't be in the UI right?

bgrenon: Acutally, I should create a new page and leave a redirect. This will appear in the URL.

pffigueiredo: I would prefer that we get read of it for once, we are really trying to start clean with "Neon Auth" - but feel free to leave it if you think that for docs that's the best approach 💪

content/docs/guides/neon-identity.md
bgrenon: Added those

content/docs/guides/neon-identity.md
bgrenon: Revised to:
"Since the `neon_auth.users_sync` table is updated asynchronously, there may be a brief delay (usually less than 1 second) before a user's data appears in the table. This means foreign key constraints could temporarily fail for new users. Consider this possible delay when deciding whether to use foreign keys in your schema."

pffigueiredo (approved):

content/docs/guides/neon-auth.md
```diff
@@ -0,0 +1,180 @@
...
+    author_id UUID REFERENCES neon_auth.users_sync(id) ON DELETE SET NULL,
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+);
+```
+
+### Querying user data
+
+When querying data that relates to users:
+
+- Use LEFT JOINs instead of INNER JOINs with the `users_sync` table in case of any sync delays. This ensures that all records from the main table (.e.g. posts) are returned even if there's no matching user in the `users_sync` table yet.
```
danieltprice, line 170:
```suggestion
- Use LEFT JOINs instead of INNER JOINs with the `users_sync` table in case of any sync delays. This ensures that all records from the main table (e.g., posts) are returned even if there's no matching user in the `users_sync` table yet.
```

danieltprice (approved):

PR #2898, opened 2025-02-07
docs: add new twin workflow guides and refactor dev test use case page

PaulieScanlon:
Hi @danieltprice / @carlotas19. This PR creates a new navigation item called Neon Twin under a new section heading called workflows. I've added two new guides for creating Neon Twins and a Twin overview page.

I've also made some changes to the dev-test use case page under docs, and a minor change to the dev-test use case page that lives at the root.


content/docs/sidebar.yaml
```diff
@@ -1070,6 +1071,20 @@
       slug: introduction/roadmap
       icon: roadmap
       tag: updated
+- section: Workflows
+  items:
+    - title: Neon Twin
+      slug: guides/neon-twin-intro
+      icon: partner-guide # TODO change icon
```
danieltprice, line 1078: we have an icon ready:

icon: workflows

PaulieScanlon: Hey @danieltprice. I've had PP create a new icon for Neon Twin. I've updated the sidebar.yaml and added the new svg.

content/docs/guides/neon-twin-full-pg-dump-restore.md
```diff
@@ -0,0 +1,166 @@
...
+
+- `name`: The name of the Action as it appears in the GitHub UI.
+- `cron`: The [POSIX cron syntax](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/crontab.html#tag_20_25_07) that defines when the Action will run.
+- `workflow_dispatch`: Enables manual triggering through the GitHub UI.
+
+### env
+
+- `PROD_DATABASE_URL`: The PostgreSQL connection string for your production database.
+- `DEV_DATABASE_URL`: The PostgreSQL connection string for your Neon database.
+- `PG_VERSION`: The version of PostgreSQL to install in the Action environment.
```
danieltprice, line 78: Do we install the Postgres instance in the Action environment to be able to run pg_dump and pg_restore commands? And I was wondering where the dump file is created. Is it created in the Action environment, and does it continued to live there? Are there any database size / dump file size limitations?

content/docs/guides/neon-twin-full-pg-dump-restore.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Testing the workflow
+
+To manually trigger your workflow go to **Actions** > **Create Neon Twin** then click **Run workflow**. From the dropdown, click the **Run workflow** button.
+
+![github actions run workflow](/docs/guides/twin_diagram_test_workflow.png)
+
+## Syncing with migration changes
+
+The GitHub Action runs on a recurring schedule, but you may also want it to trigger when migration changes are applied and a Pull Request is merged. To enable this, update the Action with the following code:
```
danieltprice, line 123: The pull request referred to here — I'm having a hard time wrapping my head around the workflow involving migrations. Are you migrating changes form the Neon Twin back to the product database and then syncing between prod and the Neon Twin?

content/docs/guides/neon-twin-full-pg-dump-restore.md
PaulieScanlon: That's correct. Postgres is installed in the "VM" a GitHub Actions Ubuntu environment. The dump file is temporarily written to disk in the same environment. When the job is complete, the environment is trashed. 

Regarding disk size, to the best of my knowledge, no. There is no mention of disk size limitations in the docs: https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#usage-limits.

content/docs/guides/neon-twin-full-pg-dump-restore.md
PaulieScanlon: Yes... well, sort of. It's hard to know how customers will be applying migrations. Ideally, the repository that is used to run migrations would also house this Action. When a PR is merged that applies the migrations is closed and merged is true, this same Action will run the steps again. This is to avoid schema drift. 

So, in short, the schedule is set to automatically sync data (as this will he constantly changing), and the schedule can be set to run as often as every 15 minutes, but the PR condition is set to sync schema.

PaulieScanlon: The `workflow_dispatch` is there so you can trigger the job manually—good while testing the Action.

PaulieScanlon: I've linked to [this blog post](https://neon.tech/blog/neon-twin-deploy-workflow) at the bottom of the guide that explains the workflow in more detail.

danieltprice (approved):

PR #2903, opened 2025-02-08
Inline SQL Editor component

m-abdelwahab:
This PR adds support for an inline SQL editor component that can be embedded on our Postgres tutorial pages.


saimonkat:
Hi @m-abdelwahab do you think this PR is still relevant?

PR #2979, opened 2025-02-20
docs: add Twin workflow diagram to use case dev test page and twin overview page

PaulieScanlon:
Add `dev-test-twin-workflow.png` to:
1. docs/use-cases/dev-test
2. docs/guides/neon-twin-intro

The image is shared between the two pages, but I've only saved it in `public/use-cases`. Should I also create a copy of the image and save it in `public/docs/guides`?


danieltprice:
Lots of other cases of shared images. I don't see an issue with doing that!

danieltprice (approved):

PR #3010, opened 2025-02-23
fix: column name

tarunxsh:
fix column name
- index is defined on `last_name` col in `customer` table


danieltprice (approved):
Thank you @tarunxsh !

Issue #3027, opened 2025-02-26
Typo: grammar mistake at PL/pgSQL section of the docs.


TheCaptain1810:
[Link for the docs](https://neon.tech/postgresql/postgresql-plpgsql/plpgsql-block-structure)

![Image](https://github.com/user-attachments/assets/8bf189df-7db8-4ac1-97d4-111014b80d73)

Here 👆, it should be "They have the same meaning", instead of "They are have the same meaning".
I'm working on fixing it. Will link the PR shortly.

PR #3028, opened 2025-02-26
typo fix: for issue #3027

TheCaptain1810:
The typo containing grammar mistake where it had a sentence like "They are have the same meaning" instead of "They have the same meaning" is fixed.

Here is the issue: https://github.com/neondatabase/website/issues/3027


danieltprice (approved):
Thank you @TheCaptain1810 !

PR #3036, opened 2025-02-26
fix: missing param in postgresql-full-text-search.md

kofta999:
Fixes the missing `regconfig` parameter in `to_tsvector()` function that results in throwing this error if not provided: `ERROR:  generation expression is not immutable`

Used `english` just like the correct usage when using GIN index.


danieltprice (approved):

danieltprice:
Thank you @kofta999

PR #3071, opened 2025-03-04
Update security-reporting.md - update private to public for bb program

areyou1or0:
context: https://neon.tech/blog/neons-bug-bounty-program-with-hackerone-goes-public


danieltprice (approved):
Thank you @areyou1or0

PR #3100, opened 2025-03-07
fix: correct import path for currentUser in Clerk

jnzlab:
Updated the import path for currentUser from @clerk/nextjs to @clerk/nextjs/server to resolve an import error. This change ensures compatibility with the latest Clerk package structure and prevents runtime issues.


danieltprice (approved):

danieltprice:
Thank you @jnzlab

PR #3228, opened 2025-03-25
docs: Fix LFC size ratio

salim-b:
I think ***75 %*** is correct, at least that's what's explained [here](https://neon.tech/docs/manage/endpoints#how-to-size-your-compute).


danieltprice (approved):

danieltprice:
Thank you @salim-b !

PR #3233, opened 2025-03-27
docs: add pg_subscription_rel and lr troubleshooting guides

dhanushreddy291:
https://neon-next-git-fork-dhanushreddy291-add-docs-247336-neondatabase.vercel.app/docs/guides/logical-replication-pg_subscription_rel


bgrenon:
@dhanushreddy291 the pgcopydb doc is reviewed with some edits, mostly for concision, but with 2 new points:
-  When I tested (neon to neon), --resume did not work. I think the source db needs to support traditional postgres snapshots. I've added language to that effect. Can you confirm this?
- The environment variables from Step 1 were not available to the 2nd monitoring terminal session; I added language capture you need to either set new variables, or include --source  flag.

dhanushreddy291:
> @dhanushreddy291 the pgcopydb doc is reviewed with some edits, mostly for concision, but with 2 new points:

Sorry, @bgrenon. I tested LR from other providers to Neon but missed Neon-to-Neon and Neon to third-party postgres. I encountered errors today when doing that and added `--no-acl` for Neon-to-Neon migrations or Neon to any.

The following error is when we do `pgcopydb clone --no-owner` from Neon to Neon without `--no-acl`
image

The following error is when we do `pgcopydb clone --no-owner` from Neon to a 3rd party Postgres without `--no-acl`
image


> * When I tested (neon to neon), --resume did not work. I think the source db needs to support traditional postgres snapshots. I've added language to that effect. Can you confirm this?

I also removed the [resume](https://github.com/dimitri/pgcopydb/issues/659#:~:text=the%20resume%20operation%3F-,My%20tests%20show%20that%20%2D%2Dresume%20only%20works%20for%20CDC%20(with%20our%20approach%20whithout%20errors%2C%20point%203.).,-Tested%20the%20pgcopydb) flag and stream sentinel since they don't work without functional LR. My apologies for the oversight.

> * The environment variables from Step 1 were not available to the 2nd monitoring terminal session; I added language capture you need to either set new variables, or include --source  flag.

Sure. Thanks

philip:
Hi @dhanushreddy291, do you have thoughts on the state of this PR? Looks like [pgcopydb](https://neon.com/docs/import/pgcopydb) was merged into the docs, is the `logical-replication-pg_subscription_rel` guide still accurate and useful?

PR #3248, opened 2025-03-28
docs: add pub/sub using listen/notify guide

vkarpov15:


content/guides/pub-sub-listen-notify.md
```diff
@@ -0,0 +1,94 @@
...
+---
+
+PostgreSQL has a built-in mechanism for publish/subscribe (Pub/Sub) communication using the LISTEN and NOTIFY commands.
+This allows different sessions to send messages to each other using Postgres, without needing a separate service like Kafka or RabbitMQ.
+
+## Steps
+
+* Overview
+* Set Up a Listener in Node.js
+* Send a Message using NOTIFY
```
danieltprice, line 17:
```suggestion
* Send a Message using `NOTIFY`
```

content/guides/pub-sub-listen-notify.md
```diff
@@ -0,0 +1,94 @@
...
+* Limitations of LISTEN/NOTIFY
```
danieltprice, line 18:
```suggestion
* Limitations of `LISTEN`/`NOTIFY`
```

content/guides/pub-sub-listen-notify.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+Note that you don't need to explicitly create a channel.
+
+### Limitations of LISTEN/NOTIFY
+
+Postgres `LISTEN` and `NOTIFY` run entirely in memory and do not persist any data.
+If there are no listeners when a `NOTIFY` runs, the message disappears and Postgres does not provide a mechanism to replay messages.
+
+Because `LISTEN` and `NOTIFY` run in memory,
```
danieltprice, line 91: this sentence is incomplete.

danieltprice (changes requested):
There's a compatibility item that should be mentioned with Neon's scale to zero behavior.

danieltprice (approved):

PR #3325, opened 2025-04-14
Fix real values range typo

SadEagle:
Fix real value range  typo
Was tricked...


danieltprice (approved):

danieltprice:
Thank you @SadEagle !

PR #3376, opened 2025-04-23
Update postgresql-row_number.md

lightify97:
ROW_NUMBER is not defined as an alias and will cause an error. Defined an alias for the column for future reference outside of the subquery.


danieltprice (approved):

danieltprice:
Thank you @lightify97

PR #3434, opened 2025-05-05
docs: branch db limit

danieltprice:


danieltprice:
addressed by another PR

PR #3452, opened 2025-05-09
feat: videos optimization

saimonkat:
This PR brings project videos optimization


saimonkat:
It’s just a small size saver, which reduces videos quality, so it doesn’t have any sense
Better to move videos originals to cdn

PR #3483, opened 2025-05-16
Update postgresql-insert-multiple-rows.md

macbobby913:
There was a typo in the sentence: "The following statement uses the INSERT statement to insert three rows into the links table:". It looks like "links" should be replaced with "contacts".


danieltprice (approved):

danieltprice:
Thank you @macbobby913

PR #3490, opened 2025-05-19
Normalize File Path Separators and Update Project Dependencies

Hussain5618:
**Fix**: Improved the getPostSlugs function to normalize file path separators for cross-platform compatibility, ensuring consistent behavior on Windows and Unix-based systems.
**Fix**: Adjusted the z-index value for the header component to resolve overlapping issues.
**Chore**: Removed the unused dependency @splinetool/react-spline from [package.json] to clean up the project.

451b0803-1eb1-41b2-ae2f-89e8d78d309a
8cbb8a86-82a0-4540-a1b2-6bba5654bd50

[Preview](https://neon-next-git-fork-hussain5618-main-neondatabase.vercel.app/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,6 @@
         "@react-hook/window-scroll": "^1.3.0",
         "@rive-app/react-canvas": "^4.9.5",
         "@segment/analytics-next": "^1.67.0",
-        "@splinetool/react-spline": "^2.2.6",
         "@tailwindcss/typography": "github:tailwindcss/typography",
         "@upstash/ratelimit": "^2.0.5",
         "algoliasearch": "^5.20.0",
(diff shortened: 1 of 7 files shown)


saimonkat:
I couldn't find a page with this header overlapping issue, can you specify please?
image

saimonkat (changes requested):
@Hussain5618 regarding the `getPostSlugs` function refactor:

The `glob` package (which is being used to get the files) already handles cross-platform path separators internally. It returns paths in a consistent format regardless of the operating system

The current implementation is actually simpler and works fine because:
- `file.replace(pathname, '')` removes the base path
- `.replace('.md', '')` removes the file extension
- The resulting string will be used as a URL path, where forward slashes are the standard

The proposed change adds an extra complexity by:
- Using `path.relative()` to get a relative path
- Using `path.posix.normalize()` to ensure forward slashes
- These steps are redundant since the paths are already in the correct format

The only potential edge case would be if the pathname parameter contained backslashes on Windows, but since this is a Next.js application (as evident from the codebase), the pathname parameter would typically be provided with forward slashes anyway.

I would recommend keeping the current implementation as it's simpler and works correctly across platforms.

PR #3540, opened 2025-06-02
docs: include headings from shared content in table of contents

bgrenon:


bgrenon:
@saimonkat can you take a look at this? 
it's vibed so feel free to alter anything as needed

bgrenon:
> So I tried a lot of things to resolve it automatically, but finished up with the manual method. Now we can specify unique `id` for headings manually using the following format:

Oh, nice! Manual is not a problem, just great to get these kind of dupe headings to render nicely.

PR #3581, opened 2025-06-19
Update import-data-assistant.md

danieltprice:
update migration assistant docs


content/docs/import/import-data-assistant.md
```diff
@@ -43,6 +43,7 @@ You'll need:
   ```
 - **Admin privileges** on your source database. We recommend using a superuser or a user with the necessary `CREATE`, `SELECT`, `INSERT`, and `REPLICATION` privileges.
 - A database **smaller than 10 GB** in size for automated import
+- When migrating from another Postgres provider, we recommend migrating to a Neon project created in the same region as your current database. This helps ensure a faster import. Please note that there is a 1-hour time limit on import operations. Import operations that exceed this limit will fail.
```
danieltprice:
```suggestion
- We recommend migrating to a Neon project created in the same region as your current database. This helps ensure a faster import. There is a 1-hour time limit on import operations.
```

PR #3628, opened 2025-07-06
Update postgresql-select-distinct.md

SefterM-zade:


philip:
Thanks for the fix @SefterM-zade!

PR #3652, opened 2025-07-11
docs: add ttl branching docs

philip:
Adds documentation for TTL branches, branches that auto-delete at specified timestamps. Includes API/CLI/Console usage, examples, and integration with existing docs.

Note: Documentation prepared for review; actual feature behavior to be validated once implemented.

[Preview](https://neon-next-g213ieq34-neondatabase.vercel.app/docs/guides/branch-expiration)


content/docs/guides/ttl-branches.md
```diff
@@ -0,0 +1,300 @@
...
+     --data '{
+       "branch": {
+         "expires_at": null
+       }
+     }'
+```
+
+```bash
+# Example 1: Update expiration to a new timestamp
+neon branches set-expires-at <branch-id> "2025-07-20T12:00:00Z" --project-id <project-id>
```
gsalomao: TBH, I think this approach quite confusing and I don't like adding a new command for each attribute. Wdyt in a command like this:
```
neon branches update <branch-id | branch-name> --expires-at="2025-07-20T12:00:00Z"
```

In this way, users could combine multiple attributes in the same command:
```
neon branches update <branch-id | branch-name> --default --expires-at="2025-07-20T12:00:00Z" --protected=false
```

content/docs/guides/ttl-branches.md
```diff
@@ -0,0 +1,300 @@
...
+
+# Example 3: Remove expiration from a branch
+neon branches set-expires-at <branch-id> null --project-id <project-id>
+```
+
+```markdown
+1. Navigate to the Branches page in the Neon Console
+2. Find the branch you want to modify
+3. Click the three-dot menu (...) next to the branch
+4. Select "Edit expiration"
```
gsalomao: @lpetkov, could you please take a look at this flow regarding update branch's expiration timestamp? Let's know if you plan something different.

content/docs/guides/ttl-branches.md
```diff
@@ -0,0 +1,300 @@
...
+</CodeTabs>
+
+## Capabilities
+
+### Managing branch expiration
+
+You can set, update, or remove expiration timestamps on branches through multiple interfaces:
+
+- **API** - Use the `expires_at` parameter with [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6) format when creating or updating branches (see [API reference](#api-reference)).
+- **CLI** - Use the `--expires-at` flag when creating branches or the `set-expires-at` subcommand to update expiration timestamps (see [timestamp format](#timestamp-format)).
```
gsalomao: Same as my previous comment about the `set-expires-at` subcommand.

content/docs/guides/ttl-branches.md
philip: I assumed `neon branches update` existed but then realized it didn't, so I made a guess that follows the other `neon branches` options like `set-default` and `add-compute`. But for what it's worth, I do love the idea of having a unified `update` option.

philip:
As @danieltprice suggested, this PR won't be merged. Instead, the attributes will be documented with other branching documentation (i.e., there's new branch functionality but not a new branch type).

philip:
The PR was updated, it now references branches that can expire instead of a unique TTL Branches feature.

content/docs/guides/expired-branches.md
```diff
@@ -0,0 +1,297 @@
...
+Branch expiration is ideal for temporary branches that have predictable lifespans:
+
+- **CI/CD environments** - Test branches that should clean up after pipeline completion
+- **Feature development** - Time-boxed feature branches with known deadlines
+- **Automated testing** - Ephemeral test environments created by scripts
+- **AI workflows** - Temporary environments managed without human intervention
+
+Without automatic expiration, these branches accumulate over time, increasing storage costs and project clutter.
+
+### Recommended expiration times
```
danieltprice: Should we add a warning here that deleted branches are not recoverable? Something like this, maybe:

"Branch deletion is permanent. Once an branch is deleted, all data on the branch is lost and cannot be recovered. Compute endpoints associated with the branch are also removed when a branch is deleted. Give careful consideration to your expiration times and double-check them to avoid data loss."

danieltprice: Also, "recommended" seems too strong. This will totally depend on user requirements, and I would not want users to blindly follow this advice. How about:

**Example branch expiration settings**  
Here are some example expiration settings that teams might use, depending on the purpose of the branch. Adjust these to fit your workflow and branch management requirements.

- CI/CD pipelines: 2–4 hours  
- Demo environments: 24–48 hours  
- Feature development: 1–7 days  
- Long-term testing: 30 days

content/docs/guides/expired-branches.md
```diff
@@ -0,0 +1,297 @@
...
+
+<Admonition type="important">
+Branch deletion is irreversible. Once deleted, a branch and its data cannot be recovered. All compute endpoints associated with the branch are also deleted. Deletion typically occurs within minutes of the expiration time.
+</Admonition>
+
+## Setting branch expiration
+
+You can set, update, or remove expiration timestamps through three interfaces:
+
+- **API** - Use the `expires_at` parameter with [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6) format
```
danieltprice: We mention ISO8601 date/time format elsewhere in the docs and API ref. Should we reference to that or the RFC? I'm not sure, but I think we should use one or the other.

content/docs/guides/expired-branches.md
```diff
@@ -0,0 +1,297 @@
...
+- **CLI** - Use the `--expires-at` flag when creating branches or the `set-expires-at` subcommand to update
+- **Neon Console** - Use the date and time selector (automatically handles formatting)
+
+See the [Examples](#examples) section below for detailed usage of each method.
+
+## Timestamp format requirements
+
+The `expires_at` parameter must use [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6) format with second-level precision:
```
danieltprice: same comment as above.

content/docs/guides/expired-branches.md
```diff
@@ -0,0 +1,297 @@
...
+- Time zone is required (either `Z` or numeric offset, not both)
+- Fractional seconds are optional but only second precision is stored
+- Timestamp must be in the future
+- Maximum expiration is 1 year from the current time
+
+<Admonition type="note">
+Common errors include missing timezone (`2025-07-15T18:02:16`), past timestamps, or combining `Z` with offset (`2025-07-15T18:02:16Z-05:00`). The API will return a validation error for invalid formats.
+</Admonition>
+
+## Understanding TTL behavior
```
danieltprice: Should we introduce this acronym somewhere before using it? e.g. time-to-live (TTL).

danieltprice: I find this section hard to understand. 
- The title is a little ambiguous
- Is it just about retrieving branch data via the API or does it apply to values I'll see in the UI as well? 
- instead of "TTL interval", should we just say "ttl_interval_seconds value" to avoid the reader having to translate it as they read
- instead of saying "preserved to ensure consistent behavior", can we say that the interval doesn't change, only the expires_at value does — or something like that (simpler language)
- I had to read it a couple of times to understand where the "24 hours" comes from (I missed it in the code comment), so maybe explain that upfront?

content/docs/guides/expired-branches.md
```diff
@@ -0,0 +1,297 @@
...
+```
+
+When you reset this branch from its parent, `expires_at` updates to the current time plus 24 hours, as 24 hours was the original interval.
+
+## Restrictions
+
+To maintain system integrity, expiration timestamps cannot be added to:
+
+- **Protected branches** - Cannot expire protected branches or protect branches with expiration
+- **Default branches** - Cannot expire default branches or set expiring branches as default
```
danieltprice: What about root branches? You can have root branches and default branches. They do not need to be the same branch.
https://neon.com/docs/manage/branches#branch-types

content/docs/guides/expired-branches.md
```diff
@@ -0,0 +1,297 @@
+---
+title: Expired branches
```
gsalomao: Are we changing the terminology from `TTL branches` to `Expired branches`? If so, we need to coordinate this because marketing materials are being created using the `TTL branches` terminology.

PR #3757, opened 2025-08-04
update projects pages

dhanushreddy291:


dhanushreddy291:
`We'll post a notice there at least **1 day** ahead of a planned update, letting you know when it's coming`

Though not in projects pages, it is in `updates`: https://neon.com/docs/manage/updates#updates-on-the-free-plan
I felt its no longer valid as UI is not showing this right now.

(I still updated the image as it was a common one and there is no mention of update announcements note)

old image

danieltprice (approved):

PR #3864, opened 2025-08-26
fix: wait for async calls to emit zaraz identify and track events

rodneyshibu:
**Context**
We want to start funneling startup form data into customer.io in addition to HubSpot. We have an existing pipeline to send events to Cloudflare Zaraz -> Segment -> Customer.IO. However, we've noticed that in some cases, only one of `identify` or `Startup Form Submitted` ends up reaching Segment.

**What changes are proposed in this pull request?**
This PR fixes the logic by awaiting the promise before emitting the next track event while also adding guardrails in case `window.zaraz.track` does not return a promise. Other existing uses of `sendGtagEvent` should work the same (i.e. this change should be backward compatible).

**How to revert?**
The zaraz integration with the startup form is behind the following PostHog feature flag: https://us.posthog.com/project/49356/feature_flags/176002 

**How is this tested?**
- Tested this locally by enabling zaraz in dev env and verifying that the events are received by Segment.
working_startup_form_submit

#[LKB-2733](https://databricks.atlassian.net/browse/LKB-2733)

### DIFF
--- a/src/utils/send-gtag-event.js
+++ b/src/utils/send-gtag-event.js
@@ -1,5 +1,7 @@
 export default function sendGtagEvent(eventName, properties) {
   if (window.zaraz) {
-    window.zaraz.track(eventName, properties);
+    const result = window.zaraz.track(eventName, properties);
+    return Promise.resolve(result);
   }
+  return Promise.resolve();
 }
--- a/src/components/pages/startups/hero/contact-form/contact-form.jsx
+++ b/src/components/pages/startups/hero/contact-form/contact-form.jsx
@@ -148,8 +148,8 @@ const ContactForm = () => {
         if (isFormDataSentToCustomerIo) {
           try {
             if (window.zaraz && email) {
-              sendGtagEvent('identify', { userId: email, email });
-              sendGtagEvent(eventName, eventProps);
+              await sendGtagEvent('identify', { email });
+              await sendGtagEvent(eventName, eventProps);
             }
           } catch (error) {
             console.warn('Error submitting the form');


src/utils/send-gtag-event.js
```diff
@@ -1,5 +1,7 @@
 export default function sendGtagEvent(eventName, properties) {
   if (window.zaraz) {
-    window.zaraz.track(eventName, properties);
+    const result = window.zaraz.track(eventName, properties);
+    return Promise.resolve(result);
```
duskpoet, line 4: why do you wrap result in a promise?

duskpoet: if you call `await` on non-promise value, it's a noop, so shouldn't be a problem

rodneyshibu: It was more to indicate that `sendGtagEvent` returns a promise so that, in the future, others will know to `await` it to avoid any potential issues. I could just add a comment instead, though.

duskpoet: ideally it shouldn't be required to await, such things should be sent in batches, that's what segment does for example. Maybe we can use segment directly?

rodneyshibu: Yup, we considered calling the endpoints directly (and I drafted this [one-pager](https://docs.google.com/document/d/1iDPiOB_aZK7b7jxyNr8j-nSnd7gvxwBwPxBCvm9t_so/edit?tab=t.0) here with the tradeoffs), but since this pipeline via Zaraz was already built out before, we decided to go with the quickest approach here. It also avoided a Databricks security review since we aren't adding any new API calls (even though the data still flows through segment). Here's the slack thread for more context: https://databricks.slack.com/archives/C091EQGTY67/p1754666579143419

duskpoet: I don't think we need any security review for segment, we already use it. 
let's see if this satisfies the cause, but it looks pretty weird to me, that we are not using semgent directly, I don't see why, but it obviously causes troubles

rodneyshibu: Yeah, agreed that using segment directly would be better. Some users have rigorous privacy settings set up, though (like Firefox Enhanced Tracking Protection) that can block cross-site calls, which is why most tracking logic on the public site goes through Cloudflare. We can probably get around that using API routes, but then we'd need to restructure the logic and test the new cases. I'll flag this later today though.

duskpoet: that's why we have segment proxy on our domain :)

PR #3977, opened 2025-09-23
Adding self-verifying jwt and remove refs into neon rls

bgrenon:
Refers to: [LKB-4819](https://databricks.atlassian.net/browse/LKB-4819)

- Adding info about self-verifying JWTs to handle the backend use case for RLS
- removing external reference into the soon-to-be-removed Neon RLS content

Preview:
- https://neon-next-git-bgrenon-self-verified-jwts-neondatabase.vercel.app/docs/serverless/serverless-driver#using-transactions-with-jwt-self-verification
- https://neon-next-git-bgrenon-self-verified-jwts-neondatabase.vercel.app/docs/guides/rls-drizzle#understanding-neons-auth-functions

### DIFF
--- a/content/docs/guides/rls-drizzle.md
+++ b/content/docs/guides/rls-drizzle.md
@@ -32,24 +32,14 @@ Drizzle ORM provides a declarative way to manage these policies directly within 
 
 ## Understanding Neon's auth functions
 
-The Neon [Data API](/docs/data-api/get-started) provides the `auth.user_id()` function that automatically extracts user information from JWT claims and makes it available in your RLS policies. It works like this:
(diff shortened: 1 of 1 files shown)


content/docs/guides/rls-drizzle.md
```diff
@@ -31,18 +30,37 @@ Row-Level Security (RLS) is an important last line of defense for protecting you
...
+```typescript
+// In your RLS policy
+using: sql`(select auth.user_id() = ${table.userId})`,
+```
+
+If you want to use your own JWT validation, you can use something like:
+
+```javascript
+// Set JWT claims and query data in a transaction
+const [, my_table] = await sql.transaction([
```
mrl5: please note that this is "raw" serverless driver and we're mentioning it in `rls-drizzle` guide. Not sure if this is a good fit here. WDYT?

PR #4120, opened 2025-11-05
docs(pages): rename `Neon Launchpad` -> `Instagres`

atilafassina:


content/docs/reference/instagres.md
```diff
@@ -89,7 +91,7 @@ npx get-db --env ./my.env --key MY_DB_URL
 # Skip prompts and use defaults
 npx get-db --yes
 
-# Detects PUBLIC_NEON_LAUNCHPAD_CLAIM_URL (default) from your environment,
+# Detects PUBLIC_NEON_INSTAGRES_CLAIM_URL (default) from your environment,
```
philip: When will this change go live?

content/docs/reference/instagres.md
```diff
@@ -89,7 +91,7 @@ npx get-db --env ./my.env --key MY_DB_URL
 # Skip prompts and use defaults
 npx get-db --yes
```
philip, line 92: When will this become instagras instead of get-db?

content/docs/reference/instagres.md
atilafassina: today :)

content/docs/reference/instagres.md
atilafassina: this may take a tiny bit longer, but even then `get-db` will still work

PR #4223, opened 2025-12-10
feat: enhance proxy and neonAuthMiddleware docs

andrelandgraf:


content/docs/auth/quick-start/nextjs.md
```diff
@@ -95,19 +95,19 @@ The `neonAuthMiddleware()` ensures that user is authenticated before the request
...
   // Redirects unauthenticated users to sign-in page
-  loginUrl: "/auth/sign-in"
-})
+  loginUrl: "/auth/sign-in",
+});
 
 export const config = {
   matcher: [
-    // Run the middleware for all paths, except the static resources
-    "/((?!_next/static|_next/image|favicon.ico).*)",
```
andrelandgraf, line 108: Doesn't include `/public` folder assets etc. which is annoying - e.g., images and other public assets fail to load with this setup.

content/docs/auth/quick-start/nextjs.md
```diff
@@ -195,8 +195,8 @@ const geistMono = Geist_Mono({
 });
 
 export const metadata: Metadata = {
-  title: 'My Neon App', // [!code ++]
-  description: 'A Next.js application with Neon Auth', // [!code ++]
```
andrelandgraf, line 199: Not relevant to highlight

content/docs/auth/quick-start/nextjs.md
```diff
@@ -95,19 +95,19 @@ The `neonAuthMiddleware()` ensures that user is authenticated before the request
...
+});
 
 export const config = {
   matcher: [
-    // Run the middleware for all paths, except the static resources
-    "/((?!_next/static|_next/image|favicon.ico).*)",
-  ]
-}
+    // Protected routes requiring authentication
+    "/account/:path*",
```
Shridhad, line 108: What is general practice with these matcher - Do users prefer to configure allow list or disallow list? 

I feel its more secure and safe to protect every route/resource and only explicitly configure exclude routes/resources that are public.

andrelandgraf: Both isn't great imo! Ideally, you should enforce auth rules on every page individually, co-located with the page code. `proxy.ts` is too easy to forget. Either you forget to allowlist something and then resources aren't loading/available or you forget to denylist something and you have an insecure page. 

I would recommend: disallow list in `proxy.ts` and additionally explicit authorization checks on every RSC page and API endpoint for fine-grained checks. 

`proxy.ts` as a quick (almost optional) cookie expiration check and then in the RSC when loading the required page data, do one more check to see if the user has access to the resources. For this, I would likely always fetch the user from the DB (full auth check) again.

andrelandgraf: But for this minimal guide, I'd go with denylist to avoid frustration when `public/` assets aren't loading etc.

Shridhad: > Both isn't great imo! Ideally, you should enforce auth rules on every page individually, co-located with the page code.
> I would recommend: disallow list in proxy.ts and additionally explicit authorization checks on every RSC page and API endpoint for fine-grained checks.

I agree. the denylist does make sense to avoid the frustration. @bgrenon  Lets put this recommendation in docs as well

> proxy.ts as a quick (almost optional) cookie expiration check and then in the RSC when loading the required page data, do one more check to see if the user has access to the resources. For this, I would likely always fetch the user from the DB (full auth check) again.

Once we implement session data (not just token) in the cookies that nextjs package can encrypt/decrypt, we can utilize React's `cache` to resolve the session data to retrieve - session details, user data, user roles, etc. To some extend we can even include user's organization details here. For everything else, it makes sense to make upstream API call like update from RSC, user/account settings, etc

danieltprice (approved):

PR #4286, opened 2025-12-25
feat: add filename to code block in docs

vannyle:
## Description

- Updated MDX code blocks to support displaying a filename via fenced code meta (`filename="..."`).
image

### Changes

- Fixed Shiki line highlighting to recognize `{...}` ranges anywhere in the code fence meta (not only as the first token).
- Quoted values work and unquoted work (`filename="..."` and `filename=...`).
- Removed `CodeWithLabel` from the MDX component map and converted all docs that used `<CodeWithLabel>` to use `filename="..."` on fenced code blocks.
- Deleted the now-unused `CodeWithLabel` component files.

### Checklist

- [x] Pixel Perfect


bgrenon (approved):
Much improved - thank you!

PR #4314, opened 2026-01-06
[chore]: update docs for SolidStart

atilafassina:
- [x] Update to use SolidStart logo instead of SolidJS
- [x] use `query()` function to deduplicate queries on code snippets
- [x] use `json()` helper to format Response
- [x] use `process.env` instead of `import.meta.env` during server runtime
- [x] replace `load()` with `preload()`


philip:
@atilafassina Do you mind updating the [associated AI prompt](https://github.com/neondatabase/website/blob/main/public/prompts/solidstart-prompt.md) as well? I can later after this PR is approved/merged, but mentioning here as a reminder.

PR #4316, opened 2026-01-06
docs: add limits page

philip:
Drafted a centralized page for usage limits. Setting to DRAFT, needs more review/research.

### DIFF
--- a/content/docs/reference/limits.md
+++ b/content/docs/reference/limits.md
@@ -7,6 +7,19 @@ updatedOn: '2026-01-06T00:00:00.000Z'
 
 Find all Neon platform limits below, organized by category. Limits vary by [plan](/docs/introduction/plans) unless otherwise noted.
 
+## Project and branch limits
+
+| Limit                           | Free | Launch | Scale | Notes                                                                 |
+| :------------------------------ | :--- | :----- | :---- | :-------------------------------------------------------------------- |
+| Projects per account            | 100  | 100    | 1,000 | Scale is a soft limit; contact Support to increase                    |
+| Branches per project (included) | 10   | 10     | 25    | Additional branches billed separately on paid plans                   |
+| Max branches per project        | 10   | 5,000  | 5,000 | Can be increased on request                                           |
+| Root branches per project       | 3    | 5      | 25    | Includes production branch, backup branches, and schema-only branches |
+| Protected branches              | N/A  | 2      | 5     | -                                                                     |
+| Snapshots                       | 1    | 10     | 10    | Point-in-time copies of root branches                                 |
+
+For more information, see [Neon plans](/docs/introduction/plans) and [Manage branches](/docs/manage/branches).
+
 ## Storage limits
 
 | Limit                           | Free                                                      | Launch | Scale | Notes                                                     |
@@ -22,7 +35,7 @@ For more information, see [Storage](/docs/introduction/plans#storage) and [Manag
 | :--------------------------- | :------------------------- | :---------------- | :----------------- | :------------------------------------------------- |
 | Concurrently active computes | 20                         | same              | same               | Default branch exempt; contact Support to increase |
 | Autoscaling max              | 2 CU (8 GB RAM)            | 16 CU (64 GB RAM) | 16 CU (64 GB RAM)  | -                                                  |
-| Fixed compute max            | 2 CU (8 GB RAM)            | 16 CU (64 GB RAM) | 56 CU (224 GB RAM) |                                                    |
+| Fixed compute max            | 2 CU (8 GB RAM)            | 16 CU (64 GB RAM) | 56 CU (224 GB RAM) | -                                                  |
 | Compute hours                | 100 CU-hours/project/month | Usage-based       | Usage-based        | Free plan limit resets monthly                     |
 
 Each Compute Unit (CU) provides approximately 4 GB of RAM. For more information, see [Compute](/docs/introduction/plans#compute) and [Manage computes](/docs/manage/computes).
@@ -31,7 +44,7 @@ Each Compute Unit (CU) provides approximately 4 GB of RAM. For more information,
 
 | Limit                          | Value                 | Notes                                         |
 | :----------------------------- | :-------------------- | :-------------------------------------------- |
-| Postgres `max_connections`     | 104–4,000             | Varies by compute size; see table below       |
+| Postgres `max_connections`     | 104–4,000             | Varies by compute size                        |
 | Pooled connections (PgBouncer) | 10,000 concurrent     | Across all connections to the pooler endpoint |
 | default_pool_size              | 0.9 × max_connections | Per user/database pair                        |
 | max_prepared_statements        | 1,000                 | PgBouncer setting                             |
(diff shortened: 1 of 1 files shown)


content/docs/reference/limits.md
```diff
@@ -0,0 +1,204 @@
+---
+title: Usage limits
+subtitle: Reference for Neon platform limits by plan
+enableTableOfContents: true
+updatedOn: '2026-01-06T00:00:00.000Z'
+---
+
+This page provides a comprehensive reference of all limits in the Neon platform, organized by category. Limits vary by [Neon plan](/docs/introduction/plans) unless otherwise noted. For [Agent plan](/docs/introduction/agent-plan) specific limits, see the [Agent plan limits](#agent-plan-limits) section.
+
+<Admonition type="tip">
```
danieltprice: Let's leave this out to avoid creating support requests.

content/docs/reference/limits.md
```diff
@@ -0,0 +1,204 @@
...
+
+For more information, see [Storage](/docs/introduction/plans#storage) and [Manage computes](/docs/manage/computes).
+
+## Compute limits
+
+| Limit                        | Free                       | Launch            | Scale              | Notes                                              |
+| :--------------------------- | :------------------------- | :---------------- | :----------------- | :------------------------------------------------- |
+| Concurrently active computes | 20                         | same              | same               | Default branch exempt; contact Support to increase |
+| Autoscaling max              | 2 CU (8 GB RAM)            | 16 CU (64 GB RAM) | 16 CU (64 GB RAM)  | -                                                  |
+| Fixed compute max            | -                          | -                 | 56 CU (224 GB RAM) | Scale plan only                                    |
```
danieltprice:
```suggestion
| Fixed compute max            | 2 CU (8 GB RAM)                          | 16 CU (64 GB RAM)                | 56 CU (224 GB RAM) |                                     |
```

content/docs/reference/limits.md
```diff
@@ -0,0 +1,204 @@
...
+| Compute hours                | 100 CU-hours/project/month | Usage-based       | Usage-based        | Free plan limit resets monthly                     |
+| Non-default branch compute   | 5 hours/month              | -                 | -                  | Free plan only; default branch exempt              |
```
danieltprice: This 5 hour limit is the legacy limit, so it doesn't apply

content/docs/reference/limits.md
```diff
@@ -0,0 +1,204 @@
...
+
+Each Compute Unit (CU) provides approximately 4 GB of RAM. For more information, see [Compute](/docs/introduction/plans#compute), [Manage computes](/docs/manage/computes), and [Non-default branch](/docs/manage/branches#non-default-branch).
+
+## Connection limits
+
+| Limit                          | Value                 | Notes                                         |
+| :----------------------------- | :-------------------- | :-------------------------------------------- |
+| max_connections                | 104–4,000             | Varies by compute size; see table below       |
```
danieltprice:
```suggestion
| Postgres `max_connections`                | 104–4,000             | Varies by compute size; see table below       |
```

danieltprice: The settings below are all PgBouncer settings. I wonder if they should have a separate table with some info about how they apply when using a pooled connection string.

content/docs/reference/limits.md
```diff
@@ -0,0 +1,204 @@
...
+| 2                 | 8        | 839             |
+| 3                 | 12       | 1,258           |
+| 4                 | 16       | 1,678           |
+| 5                 | 20       | 2,098           |
+| 6                 | 24       | 2,517           |
+| 7                 | 28       | 2,937           |
+| 8                 | 32       | 3,357           |
+| 9–56              | 36–224   | 4,000           |
+
+<Admonition type="note">
```
danieltprice: That's the formula for compute size in an autoscaling config. Please see the link for the rest of the formula (I updated it recently).

content/docs/reference/limits.md
```diff
@@ -0,0 +1,204 @@
...
+Seven connections are reserved for Neon's internal `superuser` account. For autoscaling configurations, `max_connections` is calculated based on the formula: `min(max_compute_size, 8 × min_compute_size)`. See [Parameter settings that differ by compute size](/docs/reference/compatibility#parameter-settings-that-differ-by-compute-size) for details.
+</Admonition>
+
+For more information, see [Connection pooling](/docs/connect/connection-pooling).
```
danieltprice, line 62: I think it's important here to mention more explcitly that we support up to 10000 concurrent connections with a pooled connection string.

content/docs/reference/limits.md
```diff
@@ -0,0 +1,204 @@
...
+| Databases per branch | 500   |
+| Roles per branch     | 500   |
+
+For more information, see [Manage databases](/docs/manage/databases) and [Manage roles](/docs/manage/roles).
+
+## Read replica limits
+
+| Limit                     | Free | Paid Plans | Notes                              |
+| :------------------------ | :--- | :--------- | :--------------------------------- |
+| Read replicas per project | 3    | -          | Each replica is a separate compute |
```
danieltprice: A dash here is no limit, but above a dash for protect branches means N/A, so we'll need to be more explicit I think by using "N/A" or "No limit" or something like that

content/docs/reference/limits.md
```diff
@@ -0,0 +1,204 @@
...
+| Limit                      | Free   | Launch    | Scale     | Notes                                            |
+| :------------------------- | :----- | :-------- | :-------- | :----------------------------------------------- |
+| Monthly Active Users (MAU) | 60,000 | 1,000,000 | 1,000,000 | Contact [Sales](/contact-sales) if you need more |
+
+An MAU is a unique user who authenticates at least once during a monthly billing period.
+
+For more information, see [Neon Auth](/docs/auth/overview).
+
+## Query and request size limits
+
```
danieltprice, line 137: We should be explcit that this is a Neon serverless driver limit

content/docs/reference/limits.md
```diff
@@ -0,0 +1,204 @@
...
+| Limit                       | Value | Notes                                                          |
+| :-------------------------- | :---- | :------------------------------------------------------------- |
+| HTTP query request/response | 64 MB | Maximum size for queries over HTTP using the serverless driver |
+
+For more information, see [Neon serverless driver](/docs/serverless/serverless-driver).
+
+## Configurable quotas
```
danieltprice, line 144: Not sure this fits. It's a user defined limit in a specific use case. Maybe don't include it.

content/docs/reference/limits.md
```diff
@@ -0,0 +1,204 @@
...
+| `data_transfer_bytes`  | Project | Maximum egress data transferred                               |
+| `logical_size_bytes`   | Branch  | Maximum logical size for a branch (strict limit, not monthly) |
+
+When a quota is reached, active computes for the project are suspended until the next billing period or until the quota is increased.
+
+For more information, see [Configure consumption limits](/docs/guides/consumption-limits).
+
+---
+
+## Agent plan limits
```
danieltprice, line 150: I think this might be out of scope, maybe just point to the agent plan instead

content/docs/reference/limits.md
philip: @danieltprice Remove info from here? https://neon.com/docs/manage/branches#non-default-branch

danieltprice: @philip yes! That needs to go!

content/docs/reference/limits.md
```diff
@@ -0,0 +1,154 @@
+---
+title: Usage limits
```
danieltprice, line 2: Let's change the title to "Platform limits" to avoid confusion with billing usage and add the page here:
https://neon.com/docs/manage/platform

Maybe as a "Platform limits" topic under "Projects & resources"?

PR #4355, opened 2026-01-15
Fix result value in EXTRACT() function example

AhmedYasinKUL:
Corrected the result of the example using the EXTRACT() function.

There should be 6 decades, not 60.


philip:
Thanks for fixing this @AhmedYasinKUL

PR #4360, opened 2026-01-19
Remove compute status image from documentation

danieltprice:
Removed the compute status image from the scale to zero guide.


philip:
I added an updated version :)

PR #4392, opened 2026-01-26
Andy/remove algolia

ruf-io:


src/components/shared/blog-search/blog-search.jsx
```diff
@@ -0,0 +1,89 @@
...
+        if (value) {
+          params.set('query', value);
+        } else {
+          params.delete('query');
+        }
+        const queryString = params.toString();
+        router.replace(`${pathname}${queryString ? `?${queryString}` : ''}`, { scroll: false });
+      }, 100),
+    [pathname, router, searchParams]
+  );
```
vburlak, line 51: 1. The `debouncedUpdateURL` memo includes `searchParams` in its dependency array. `searchParams` from `useSearchParams()` returns a new object reference on every render, even if the values are unchanged. This causes the debounced function to be recreated on every render, resetting the debounce timer and breaking debouncing.

**Impact:**

- Debouncing doesn't work; URL updates happen immediately on each keystroke
- Performance impact from constant function recreation
- Unnecessary router calls

2. As well, we need to use a `router.push` instead of `router.replace` for proper browser back/forward navigation:

**Complete suggestion:**

```suggestion
  const debouncedUpdateURL = useMemo(
    () =>
      debounce((value) => {
        const params = new URLSearchParams(window.location.search);
        if (value) {
          params.set('query', value);
        } else {
          params.delete('query');
        }
        const queryString = params.toString();
        const newUrl = `${pathname}${queryString ? `?${queryString}` : ''}`;
        router.push(newUrl, { scroll: false });
      }, 100),
    [pathname, router]
  );
```

src/components/shared/blog-search/search-input.jsx
```diff
@@ -0,0 +1,54 @@
...
+  }, []);
+
+  return (
+    <div className={clsx('absolute right-0 top-1 md:w-full', className)}>
+      <div className="ais-SearchBox">
+        <input
+          ref={inputRef}
+          className={clsx('ais-SearchBox-input', asPlaceholder && 'pointer-events-none')}
+          disabled={asPlaceholder}
+          placeholder="Search..."
```
vburlak, line 38: The search input lacks an `aria-label` or `aria-labelledby`

```suggestion
          placeholder="Search..."
          aria-label="Search blog posts"
```

src/components/shared/blog-search/search-results.jsx
```diff
@@ -0,0 +1,45 @@
...
+  // Show "no results" message
+  if (posts.length === 0) {
+    return <div className="w-full text-center text-lg">No search results found</div>;
+  }
+
+  // Show filtered results
+  return (
+    <div className={clsx('search-results', className)}>
+      {posts.map((post, index) => (
+        <BlogGridItem key={post.slug} post={post} isPriority={index < 5} />
```
vburlak: When search results change, screen readers aren't notified. Dynamic content updates should use an `aria-live` region to announce changes to assistive technologies

Nice to have an `aria-live` region that announces result counts for screen readers:

```suggestion
        <>
          <div role="status" aria-live="polite" className="sr-only">
            {posts.length} results found
          </div>
          <BlogGridItem key={post.slug} post={post} isPriority={index < 5} />
        </>
```

src/components/shared/blog-search/blog-search.jsx
```diff
@@ -0,0 +1,89 @@
...
+
```
vburlak, line 55: The debounced function isn't cancelled on unmount, which can cause memory leaks and state updates after the component unmounts.

Let's add cleanup:

```suggestion
useEffect(() => {
  return () => {
    debouncedUpdateURL.cancel?.();
  };
}, [debouncedUpdateURL]);

```

src/components/shared/blog-search/blog-search.jsx
```diff
@@ -0,0 +1,89 @@
...
+import debounce from 'utils/debounce';
+
+import SearchInput from './search-input';
+import SearchResults from './search-results';
+
+const BlogSearch = ({ children, posts, searchInputClassName }) => {
+  const router = useRouter();
+  const pathname = usePathname();
+  const searchParams = useSearchParams();
+  const [query, setQuery] = useState(searchParams.get('query') || '');
```
vburlak: The `query` state is initialized from `searchParams.get('query')` but never updates when the URL changes externally (browser back/forward or external navigation). The input shows stale state.



```suggestion
  const urlQuery = searchParams.get('query') || '';
  const [query, setQuery] = useState(urlQuery);
  
  useEffect(() => {
    setQuery(urlQuery);
  }, [urlQuery]);
```

PR #4399, opened 2026-01-27
docs: migrate AI Rules to Agent Skills

philip:
Transition documentation from ai-rules (.mdc) to agent-skills format. Updates navigation, internal links, and related pages.

Note: This PR is not ready to merge but is okay to review. Thinking of alternatives to our ExternalCode usage here as it can be overwhelming. One idea is to have copy/view skill options rather than insert the entire skill into the doc.

Post-MVP enhancements include integrating skills with our custom prompts, include skill references in llms.txt, and investigate adding an llms-full.txt.


content/docs/ai/agent-skills-neon-api.md
```diff
@@ -0,0 +1,32 @@
...
+title: 'Agent Skills: Neon API'
+subtitle: Context and guidance for AI tools to use the Neon API
+enableTableOfContents: true
+updatedOn: '2026-01-27T00:00:00.000Z'
+redirectFrom:
+  - /docs/ai/ai-rules-neon-api
+---
+
+<Admonition type="note" title="Agent Skills">
+Agent Skills are new and actively being improved. We'd love your feedback! Join us on [Discord](https://discord.gg/92vNTzKDGp) to share suggestions.
```
andrelandgraf, line 11: You could add that we have an agent experience channel for this!

content/docs/ai/agent-skills-neon-api.md
```diff
@@ -0,0 +1,32 @@
+---
```
andrelandgraf, line 1: If I understand this correctly, these ai/agent-skills-* files would be indexes for a given topic from which the agent can then go deeper by following the references listed here?

content/docs/ai/agent-skills.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Example: Agent Skills in action
+
+Here's a practical example using [Cursor](https://www.cursor.so). A developer wants to set up Neon Auth in their Next.js app:
+
+**Developer prompt**: _"Set up Neon Auth for my Next.js app"_
+
+The AI assistant automatically uses the installed Neon skill to provide accurate, up-to-date guidance for authentication setup, including proper imports, configuration, and best practices.
+
+### Alternative: Browse individual references
```
andrelandgraf, line 53: I would cut this section - not sure why this would be needed :)

content/docs/shared-content/ai-skill-usage.md
```diff
@@ -0,0 +1,28 @@
...
+
+This works with:
+
+- **Cursor** - Skills are auto-discovered and invokable via slash commands
+- **Claude Code** - Skills load automatically when relevant
+- **Other agents** - Any tool supporting the Agent Skills format
+
+Once installed, your AI assistant can access {name} guidance automatically without manual configuration.
+
+### Alternatives
```
andrelandgraf, line 21: Same here - I don't think I understand the workflow one would use when following this section?

content/docs/ai/agent-skills-neon-api.md
philip: Yes, although this is a good question. Honestly, it's repurposing the existing rules pages (rename->update) so is limited in nature (only 6) for this initial conversion.

Upon reimagining, let's discuss soon about how agent skills can integrate more closely (and become) the documentation.

PR #4455, opened 2026-02-10
chore: fix typo

this-fifo:
correct `Noen` with `Neon`


philip:
Ha, thanks for the fix @this-fifo :)

PR #4513, opened 2026-02-19
docs: add neon auth organization plugin configuration docs

philip:
[preview](https://neon-next-git-neon-auth-org-plugin-neondatabase.vercel.app/docs/auth/guides/plugins/organization)


content/docs/auth/guides/plugins/organization.md
```diff
@@ -11,19 +11,122 @@ updatedOn: '2026-02-15T20:51:54.039Z'
 
 <FeatureBetaProps feature_name="Neon Auth with Better Auth" />
 
-Neon Auth is built on [Better Auth](https://www.better-auth.com/) and provides support for Organization plugin APIs through the Neon SDK. You do not need to manually install or configure the Better Auth Organization plugin.
-
-The Organization plugin allows you to build multi-tenant applications where users can create workspaces, invite other members, and manage permissions using roles.
+Neon Auth is built on [Better Auth](https://www.better-auth.com/). The Organization plugin is the way to add multi-tenancy in Better Auth. You do not need to manually install or configure the Better Auth Organization plugin.
```
Shridhad: what do you think of 

Neon Auth is built on [Better Auth](https://www.better-auth.com/) and comes with pre-configured organization plugin, so your application can support multi-tenancy from day one; no additional setup needed.

content/docs/auth/guides/plugins/organization.md
```diff
@@ -11,19 +11,122 @@ updatedOn: '2026-02-15T20:51:54.039Z'
...
 
 <Admonition type="note" title="Preview Feature">
-The Organization plugin is currently in **Beta**. Support for invitation emails and JWT token claims is currently in progress.
+The Organization plugin is currently in **Beta**. Sending invitation emails and supporting JWT token claims are currently in progress.
```
Shridhad: I would say 

`Support for invitation emails and JWT token claims is in under development.` 

OR 

`Support for invitation emails and JWT token claims will be added soon.`

content/docs/auth/guides/plugins/organization.md
```diff
@@ -11,19 +11,122 @@ updatedOn: '2026-02-15T20:51:54.039Z'
...
+Better Auth also has a **Teams** feature (sub-groups within an org); that feature is not currently enabled in Neon Auth.
+
 ## Prerequisites
 
 - A Neon project with **Auth enabled**
 - A signed-in user (organizations are associated with users)
 
+## Configure the organization plugin
+
+The Organization plugin is on by default for each branch; you can turn it off or change settings in the Console or via the API. If the plugin is disabled for a branch, your app users cannot call organization APIs (create org, invite members, and so on).
```
Shridhad: `is on by default` => `is enabled by default`

`you can turn it off` => `you can disabled it`


> If the plugin is disabled for a branch, your app users cannot call organization APIs (create org, invite members, and so on).

If the plugin is disabled, your application users and admins can not create or manage organizations, and any organization related API calls will return error

content/docs/auth/guides/plugins/organization.md
```diff
@@ -11,19 +11,122 @@ updatedOn: '2026-02-15T20:51:54.039Z'
...
+<TabItem>
+
+Open your project in the Neon Console, then go to **Auth** > **Configuration** > **Organizations** (per branch). From there you can customize:
+
+![Auth Configuration > Organizations in the Neon Console](/docs/auth/console-auth-organizations-config.png)
+
+- **Enable Organizations** (toggle): Turn the plugin on or off for the branch.
+- **Limit:** Maximum number of organizations a user can create or belong to (e.g., 1 per user).
+- **Creator role:** Role assigned to the user who creates an organization: **Owner** or **Admin**. Choose Admin if you want the org creator to have fewer privileges than Owner (for example, they cannot delete the org or change the owner).
+- **Allow user to create organization:** When on, any user can create organizations (up to the limit). When off, only users who already have Owner or Admin (the role you set in **Creator role** above) can create organizations.
```
Shridhad: > When off, only users who already have Owner or Admin (the role you set in **Creator role** above) can create organizations.

This is not right. 


Andras, can you confirm who can actually create organization when `allowUserToCreateOrganization=false`?   Maybe admin users (`user.role=admin` ) can do.  (`admin` users in the application, not in any org)

If it behaves same as `organization.enabled=false`, maybe we can remove customizing this field.

content/docs/auth/guides/plugins.md
```diff
@@ -28,7 +28,7 @@ The following Better Auth plugins are currently supported in Neon Auth:
 | [Admin](/docs/auth/guides/plugins/admin)               | ✅ Supported                                                 |
 | [Email OTP](/docs/auth/guides/plugins/email-otp)       | ✅ Supported                                                 |
 | [JWT](/docs/auth/guides/plugins/jwt)                   | ✅ Supported                                                 |
-| [Organization](/docs/auth/guides/plugins/organization) | ⚠️ Partial (invitation emails, JWT token claims in progress) |
+| [Organization](/docs/auth/guides/plugins/organization) | ⚠️ Partial (enable/configure per branch; sending invitation emails, JWT token claims in progress) |
```
Shridhad: I guess we can remove the "per branch" from here. not to cause any confusion.

content/docs/auth/guides/plugins/organization.md
philip: Thank you!

Looking at Better Auth closer [crud-org.ts#L126](https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/plugins/organization/routes/crud-org.ts#L126), `allowUserToCreateOrganization = false` blocks all session-based (client-side) creation, no role check is involved. Sessionless server-side calls (passing userId directly in the body without a session) still works.

Updated the doc to: "When off, users cannot create organizations from the client. Server-side creation is still possible."

Sound good?

thekauer: Alright so I've looked into this. Basically when that switch is turned off the only way to create orgs is if you create them on the server side by directly inserting a new org into the db. There is no way to do this using better-auth/neon-auth you have to write that code yourself.

Also the way `Limit` works in better-auth and therefore in neon as well is a little bit weird. Here is an example. If limit is set to `2` and i create an org. Then I also accept an invite to join an org. Now I can see two orgs however only one is mine. One would expect that I can still create another org but the reality is that the limit is actually based on how many orgs I am a member of. So even thought I only created one org I cannot create a new org in this situation

PR #4533, opened 2026-02-21
Fix: Updated output example in managers table

Arul-1911:
The 'type' column was removed from the output since our managers table schema does not contain it.


danieltprice (approved):
Thank you @Arul-1911

PR #4674, opened 2026-03-25
docs: remove azure integration docs

danieltprice:


next.config.js
```diff
@@ -574,6 +574,78 @@ const defaultConfig = {
         destination: 'https://get.neon.com/student-25',
         permanent: false,
       },
+      // Deprecated Neon Azure Native ISV docs (EOL); redirect to docs home
```
philip, line 577: Use our redirectFrom markdown mechanism instead?

danieltprice: The assistant recommends config for deleted or non-doc URLs. If it matters, we can revert.

PR #4713, opened 2026-04-02
Update connect-postgres-gui.md

guillim:
Adding a new postgres client to complete the list of GUI


philip:
Doesn't seem real (no source), very little info, closing until it becomes more established.

PR #4747, opened 2026-04-14
Fix next page link in the PostgreSQL EXCEPT tutorial

duffuniverse:
### Detail

Fix next page link in the PostgreSQL EXCEPT tutorial.


ruf-io (approved):
Thanks @duffuniverse

PR #4751, opened 2026-04-15
Update phrasing for DELETE JOIN explanation

rahulrao0209:
Clarified wording regarding deleting data based on another table.


ruf-io (approved):
These are great - @rahulrao0209 please email me at andy@neon.tech with your Neon account info and I will add some credit as a token of thanks

PR #4752, opened 2026-04-15
Fix repeated phrase in PostgreSQL DELETE tutorial

rahulrao0209:
Corrected a repeated phrase in the explanation of deleting multiple rows.


ruf-io (approved):
Another one! Thanks

PR #4758, opened 2026-04-16
Correct table name from 'members' to 'member'

rahulrao0209:
Updated references to 'members' table to 'member' for consistency. 
Replace contacts with member as contacts table doesn't exist in the context of this tutorial.


ruf-io (approved):
Thanks @rahulrao0209

PR #4774, opened 2026-04-18
Fix typo in tutorial on finding duplicate rows

rahulrao0209:


danieltprice (approved):

ruf-io (approved):
Correct typo fix: "The find" should be "To find". Verified against the surrounding step. Thanks!

danieltprice:
Thanks for catching this! The same fix (`The find` → `To find`) is already on `main`, so this change is now redundant. Closing as already resolved. We appreciate the contribution.

PR #4775, opened 2026-04-18
fix: correct ai skill neon-cli command syntax

sam-harri:
The neon-cli.md skill doc uses flags that don't exist. The CLI silently ignores them then falls back to the default branch so you end up running these commands against the wrong branch without knowing.

- connection-string branch should be a positional arg, not --branch-id : [docs](https://neon.com/docs/reference/cli-connection-string#the-connection-string-command)
- databases list should use --branch, not --branch-id : [docs](https://neon.com/docs/reference/cli-databases#list)
- roles list should use --branch, not --branch-id : [docs](https://neon.com/docs/reference/cli-roles#list)

Related : [neondatabase/agent-skills#21](https://github.com/neondatabase/agent-skills/issues/21)


philip:
Thanks @sam-harri, merged!

PR #4777, opened 2026-04-19
Update before-truncate-trigger.md

pradeep-lit:
NOTICE doesnt not blocks the operation


danieltprice:
Addressed both Copilot review comments in a241f5c0:
- Removed the unreachable `RETURN NULL;` after `RAISE EXCEPTION`.
- Relabeled the error output section from "Output:" to "Error:", matching the convention in the sibling `before-delete-trigger.md` tutorial.

The fix itself is correct and still needed: the tutorial's `BEFORE TRUNCATE` trigger is statement-level, where `RETURN NULL` is ignored, so only `RAISE EXCEPTION` actually aborts the truncate (the prior `RAISE NOTICE` let it through despite the prose claiming otherwise).

danieltprice (approved):

PR #4790, opened 2026-04-21
fix(guides): use custom role credentials in SST API example

ap-justin:
## Summary
- The "Building an API" section of the SST guide previously ignored the custom role and database from the "Manage Neon resources" section, using `project.connectionUri` (default role + default `neondb` database) instead
- Updated the API example to create a branch, endpoint, role (`application_user`), and database (`service_database`), then build the connection URL manually from those credentials
- Updated the explanation text and deploy output to reflect the new resources

## Test plan
- [ ] Verify the guide renders correctly
- [ ] Confirm the code examples are consistent between the two sections

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/content/guides/neon-sst.md
+++ b/content/guides/neon-sst.md
@@ -232,7 +232,7 @@ NEON_ORG_ID="<YOUR_NEON_ORG_ID>"
 
 ### Define the resources
 
-Add the following code to your `sst.config.ts` file. This code creates a Neon project and makes the connection string available to the API using resource linking.
+Add the following code to your `sst.config.ts` file. This code creates a Neon project with a custom role and database, then makes the connection string available to the API using resource linking.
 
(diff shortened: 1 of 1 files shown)


danieltprice:
Thanks for this, and for the clear rationale about the default role/database. Since this PR was opened, the SST guide was substantially rewritten and now includes a dedicated "Manage Neon resources" section that provisions a custom role (`application_user` via `neon.Role`) and database (`service_database` via `neon.Database`) — which is the pattern this PR was advocating for. Because of that rewrite, this PR no longer applies cleanly (it targets the old single-example structure).

The separate "Building an API" example still uses `connectionUri` as a deliberately simple end-to-end path, with the custom-role/database pattern documented alongside it. So we're closing this as superseded by the rewrite. Really appreciate the contribution and the reasoning.

PR #4821, opened 2026-04-26
Add PostgreSQL 19 series

bobbyonmagic:
Adding a PostgreSQL 19 series following the existing PG18 layout, eg. an overview page plus 12 feature guides covering `SQL`/`PGQ`, `ON CONFLICT DO SELECT`, `FOR PORTION OF`, `pg_plan_advice`, `REPACK`, parallel autovacuum, `JSON COPY TO`, query writing improvements, schema management, monitoring, logical replication, and breaking changes.

Examples have been run end-to-end against a master-built Postgres.

### DIFF
--- a/content/postgresql/postgresql-19/breaking-changes.md
+++ b/content/postgresql/postgresql-19/breaking-changes.md
@@ -1,0 +1,274 @@
+---
+title: 'PostgreSQL 19 Breaking Changes and Upgrade Notes'
+page_title: 'PostgreSQL 19 Breaking Changes - What to Check Before Upgrading'
+page_description: 'Review PostgreSQL 19 breaking changes including JIT disabled by default, LZ4 TOAST compression, RADIUS removal, and string handling changes before upgrading your database.'
+ogImage: ''
+updatedOn: '2026-04-15T00:00:00+00:00'
+enableTableOfContents: true
+previousLink:
+  title: 'PostgreSQL 19 Monitoring and Operations'
+  slug: 'postgresql-19/monitoring-operations'
+nextLink:
+  title: 'PostgreSQL 19 New Features'
+  slug: 'postgresql-19-new-features'
+---
+
+**Summary**: PostgreSQL 19 includes several breaking changes that may affect your applications and configuration. JIT compilation is now off by default, TOAST compression defaults to LZ4, RADIUS authentication is removed, and string handling is stricter. Review these changes before upgrading.
+
+## JIT Compilation Disabled by Default
+
+The `jit` parameter now defaults to `off`. Since PostgreSQL 12, JIT was enabled by default, but experience showed it added latency to short OLTP queries without meaningful benefit for most workloads.
+
+```sql
+-- Check your current setting
+SHOW jit;
+
+-- Re-enable if your workload benefits from it
+ALTER SYSTEM SET jit = on;
+SELECT pg_reload_conf();
+```
+
+### Who Is Affected
+
+If you run analytical or OLAP workloads with complex queries, large sequential scans, or heavy aggregations, you may see regressions after upgrading. These workloads genuinely benefit from JIT compilation.
+
+If you run typical OLTP workloads (short queries, many transactions per second), you will likely see no change or a slight improvement from avoiding JIT overhead.
+
+### What to Do
+
+Test your most important queries with `jit = on` and `jit = off` before upgrading. If JIT helps, add `jit = on` to your `postgresql.conf` explicitly. If you are unsure, leave it off - that is the safer default.
+
+## Default TOAST Compression Changed to LZ4
+
+The `default_toast_compression` parameter changed from `pglz` to `lz4`.
+
+```sql
+-- Check your current setting
+SHOW default_toast_compression;
+
+-- Revert to old behavior if needed
+ALTER SYSTEM SET default_toast_compression = pglz;
+```
+
+### What This Means
+
+LZ4 is faster than pglz for both compression and decompression, with slightly lower compression ratios. For most workloads, this is a net improvement - less CPU time spent on compression with only marginally more disk usage.
+
+Existing data is not affected. Rows already stored with pglz remain readable regardless of this setting. Only newly written or updated TOAST values use the new default. You can also override this per-column:
+
+```sql
+ALTER TABLE my_table ALTER COLUMN big_text SET COMPRESSION pglz;
+```
+
+<Admonition type="note">
+LZ4 support requires PostgreSQL to be built with `--with-lz4`. Most packaged builds include this, but verify if you build from source.
+</Admonition>
+
+## RADIUS Authentication Removed
+
+The `radius` authentication method in `pg_hba.conf` has been removed entirely. PostgreSQL will refuse to start if your `pg_hba.conf` contains any `radius` lines.
+
+### Why It Was Removed
+
+RADIUS authentication in PostgreSQL used UDP only, with no support for RADIUS over TLS (RadSec). The implementation was considered unfixably insecure for database authentication.
+
+### Migration Path
+
+Check your `pg_hba.conf` before upgrading:
+
+```bash
+grep -i radius pg_hba.conf
+```
+
+If you find RADIUS entries, switch to one of these alternatives:
+
+- **LDAP**: Most RADIUS deployments have an LDAP backend. Connect to it directly.
+- **GSSAPI/Kerberos**: For enterprise environments with Active Directory.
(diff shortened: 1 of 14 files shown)


content/postgresql/postgresql-19/breaking-changes.md
```diff
@@ -0,0 +1,274 @@
...
+enableTableOfContents: true
+previousLink:
+  title: 'PostgreSQL 19 Monitoring and Operations'
+  slug: 'postgresql-19/monitoring-operations'
+nextLink:
+  title: 'PostgreSQL 19 New Features'
+  slug: 'postgresql-19-new-features'
+---
+
+**Summary**: PostgreSQL 19 includes several breaking changes that may affect your applications and configuration. JIT compilation is now off by default, TOAST compression defaults to LZ4, RADIUS authentication is removed, and string handling is stricter. Review these changes before upgrading.
```
jkatz: I'd put the MD5 change up here. There's still a lot of usage of that - and while it won't break, the warnign behavior is a change. RADIUS has almost no usage.

content/postgresql/postgresql-19/breaking-changes.md
```diff
@@ -0,0 +1,274 @@
...
+```
+
+### What This Means
+
+LZ4 is faster than pglz for both compression and decompression, with slightly lower compression ratios. For most workloads, this is a net improvement - less CPU time spent on compression with only marginally more disk usage.
+
+Existing data is not affected. Rows already stored with pglz remain readable regardless of this setting. Only newly written or updated TOAST values use the new default. You can also override this per-column:
+
+```sql
+ALTER TABLE my_table ALTER COLUMN big_text SET COMPRESSION pglz;
```
jkatz, line 60: I'd show the general setting first (`ALTER SYSTEM SET default_toast_compression ... `).

The one thing about modifying it on a per table basis is that if you already have data in that table, you'll have to do a `VACUUM FULL` or `CLUSTER` to rewrite the storage (or if it sticks...`REPACK CONCURRENTLY` :)

content/postgresql/postgresql-19/breaking-changes.md
```diff
@@ -0,0 +1,274 @@
...
+-- 1. Change the default password encryption
+ALTER SYSTEM SET password_encryption = 'scram-sha-256';
+SELECT pg_reload_conf();
+
+-- 2. Have users reset their passwords (re-hashes as SCRAM)
+ALTER USER myuser PASSWORD 'new_password';
+
+-- 3. Update pg_hba.conf entries from md5 to scram-sha-256
+-- Change: host all all 0.0.0.0/0 md5
+-- To:     host all all 0.0.0.0/0 scram-sha-256
```
jkatz: As part of the migration step, you can leave the `md5` entry until all the passworsd are re-hashed. Postgres knows to use SCRAM if you have a SCRAM-hashed password. After re-hashing the passwords, you can then flip the HBA entry to `scram-sha-256`.

If you have un-SCRAMed passwords, `scram-sha-256` will cause them to fail.

content/postgresql/postgresql-19/breaking-changes.md
```diff
@@ -0,0 +1,274 @@
...
+
+The migration is a three-step sequence: switch the default password hash, force users to set new passwords so they re-hash under SCRAM, then flip the `pg_hba.conf` entries.
+
+```sql
+-- 1. Change the default password encryption
+ALTER SYSTEM SET password_encryption = 'scram-sha-256';
+SELECT pg_reload_conf();
+
+-- 2. Have users reset their passwords (re-hashes as SCRAM)
+ALTER USER myuser PASSWORD 'new_password';
```
jkatz: The more secure way to do this is through `psql` and `\password`, which will pre-hash the password. The above way can leak the plaintext to the log.

content/postgresql/postgresql-19/json-copy-to.md
```diff
@@ -0,0 +1,240 @@
...
+
+Generate JSON fixtures for API testing or development environments:
+
+```sql
+COPY (
+    SELECT id, name, email, role
+    FROM users
+    WHERE role = 'admin'
+    ORDER BY name
+) TO '/tmp/admin_users.json' WITH (FORMAT JSON, FORCE_ARRAY);
```
jkatz, line 192: note that copying to a file requires Postgres SUPERUSER. `\copy` will automatically handle the `TO STDOUT` and copy it to a local file: https://www.postgresql.org/docs/current/app-psql.html#APP-PSQL-META-COMMANDS-COPY

content/postgresql/postgresql-19/logical-replication-improvements.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Introduction
+
+Logical replication has been improving steadily with each PostgreSQL release. [PostgreSQL 18](/postgresql/postgresql-18/logical-replication-improvements) added retain_dead_tuples conflict-resolution support and origin-aware filtering. PostgreSQL 19 continues this trend with changes that address common operational pain points: keeping sequences in sync across replicas, managing large publications more easily, and reducing the overhead of WAL configuration.
+
+## Sequence Synchronization
+
+One of the most requested features for logical replication has been sequence synchronization. In previous versions, logical replication handled table data but ignored sequences entirely. If you used serial columns or identity columns, the sequence values on the subscriber would not advance to match the publisher, causing conflicts if the subscriber was promoted to primary.
+
+PostgreSQL 19 adds sequence synchronization to logical replication subscriptions. Subscribers can now periodically sync sequence values from the publisher.
```
jkatz, line 26: This is good particularly for using logical replication to facilitate online major version upgrades.

content/postgresql/postgresql-19/logical-replication-improvements.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Dynamic WAL Level
+
+PostgreSQL 19 introduces the `effective_wal_level` parameter, which adjusts automatically based on whether logical replication slots exist:
+
+```sql
+SHOW effective_wal_level;
+```
+
+This returns either `replica` or `logical` depending on the current state of replication slots. The system automatically adjusts the WAL level without requiring a configuration change or server restart.
```
jkatz, line 92: This is a very big deal 😄

content/postgresql/postgresql-19/logical-replication-improvements.md
```diff
@@ -0,0 +1,191 @@
...
+-- Before: required a subquery
+COPY (SELECT * FROM partitioned_sales) TO '/tmp/sales.csv' WITH (FORMAT csv);
+
+-- PostgreSQL 19: works directly
+COPY partitioned_sales TO '/tmp/sales.csv' WITH (FORMAT csv);
+```
+
+This is about 7-8% faster than the subquery approach because it avoids the overhead of query processing.
+
+## WAL Monitoring Improvements
```
jkatz: This section is for replication in general, and esp. physical replication - I'd recommend putting this into monitoring operations.

content/postgresql/postgresql-19/logical-replication-improvements.md
```diff
@@ -0,0 +1,191 @@
...
+The `pg_stat_wal` view gains a `wal_fpi_bytes` column that tracks bytes used by full-page images:
+
+```sql
+SELECT wal_records, wal_fpi, wal_fpi_bytes, wal_bytes
+FROM pg_stat_wal;
+```
+
+Full-page images can significantly increase WAL volume. Tracking their size separately helps diagnose WAL bloat.
+
+### Vacuum Progress Details
```
jkatz: This is for vacuum - this is independent of replication. I'd put this in monitoring operations

jkatz (commented):
Got through monitoring-operations.md.

One suggestion - I 'd suggest writing up the `WAIT FOR` command - this is the "read-your-writes" feature which helps to do consistent reads from async replicas: https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=447aae13b

bobbyonmagic:
Thanks for the review @jkatz! All addressed in the latest commit!

Happy to iterate if anything still feels off.

content/postgresql/postgresql-19/breaking-changes.md
ruf-io:
```suggestion
**Summary**: PostgreSQL 19 includes several breaking changes that may affect your applications and configuration. JIT compilation is now off by default, TOAST compression defaults to LZ4, MD5-hashed passwords are being deprecated and string handling is stricter. Review these changes before upgrading.
```

ruf-io (approved):

PR #4824, opened 2026-04-28
fix(docs): correct getSession example to use data.user.email

EdanStasiuk:
**Related Issue: https://github.com/neondatabase/neon/issues/12846**

## Summary

Fixes incorrect code example in the `getSession()` docs. The example was accessing `data.session.user.email`, but user is a top-level sibling of session on the returned data object, not nested inside it, as explained in https://github.com/neondatabase/neon/issues/12846#issuecomment-4331663708. Corrected to `data.user.email`.

## Tests

Verified update via local host.

### DIFF
--- a/content/docs/reference/javascript-sdk.md
+++ b/content/docs/reference/javascript-sdk.md
@@ -286,7 +286,7 @@ console.error('Sign out error:', error.message)
 const { data, error } = await client.auth.getSession()
 
 if (data.session) {
-console.log('User is logged in:', data.session.user.email)
+console.log('User is logged in:', data.user.email)
 } else {
 console.log('No active session')
 }


danieltprice:
Confirmed correct and merging. This SDK is Managed Better Auth (`@neondatabase/neon-js`), whose `getSession()` returns `{ session, user }` with `user` as a top-level sibling of `session` — not nested under it.

Verified two ways:
1. **SDK types**: better-auth types the get-session payload as `{ session, user }`.
2. **Live test**: provisioned a real Neon Auth instance and ran signUp → signIn → getSession. The response had the email at `data.user.email`, while `data.session.user` was `undefined` (the session record has no nested user).

So the original `data.session.user.email` would throw at runtime, and your `data.user.email` is right. Thanks for the fix!

danieltprice (approved):
Verified correct via SDK types and a live Neon Auth signUp→signIn→getSession test: email is at data.user.email; data.session.user is undefined.

PR #4825, opened 2026-04-28
fix: Change table name from actual_sales to sales_stats for NTILE() tutorial

FadyMorris:
This pull request makes a minor fix to the SQL example in the `content/postgresql/window-function/ntile-function.md` documentation. The change sets the correct table name used in the `SELECT` statement.

- Changed the table referenced in the SQL example from `actual_sales` to `sales_stats` in the documentation.

### DIFF
--- a/content/postgresql/window-function/ntile-function.md
+++ b/content/postgresql/window-function/ntile-function.md
@@ -70,7 +70,7 @@ SELECT
 	name,
 	amount
 FROM
-	actual_sales
+	sales_stats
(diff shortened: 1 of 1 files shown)


danieltprice:
Thanks for catching this and for the fix! The same correction (`actual_sales` → `sales_stats` in the NTILE() tutorial) landed in #5348, which has now been merged, so this table-name issue is already resolved on `main`. Closing as a duplicate. We appreciate the contribution.

PR #4826, opened 2026-04-28
fix: Change table name from actual_sales to sales_stats in PERCENT_RANK

FadyMorris:
This pull request makes a minor fix to the SQL example in the `percent_rank-function.md` documentation. The change updates the table name in the `SELECT` statement from `actual_sales` to `sales_stats` for consistency or correctness.


danieltprice:
Thanks for catching this! The same fix (`actual_sales` → `sales_stats` in the PERCENT_RANK example) is already on `main`, so this change is now redundant. Closing as already resolved. We appreciate the contribution.

PR #4828, opened 2026-04-28
Fix next page link in the PostgreSQL Foreign Key tutorial

duffuniverse:
### Detail

Fix next page link in the PostgreSQL Foreign Key tutorial


philip (approved):
LGTM — accurate navigation fix, nextLink was pointing to the wrong page.

PR #4829, opened 2026-04-28
Document storage quota write throttling behavior

bgrenon:
When a branch reaches the 16 TB platform storage limit on paid plans, writes are now throttled rather than hard-blocked (per hadron PR #5885).

Add the user-visible behavior to the 5 places where the 16 TB cap is documented, plus a clarifying paragraph in consumption-limits.md to distinguish the platform limit (slows writes) from customer-set logical_size_bytes quotas (which remain hard limits).

LKB-11977


philip (approved):
Ship it!

PR #4839, opened 2026-05-01
fix(docs) : Fixing the environment variable & Vercel Deployment Build command

ayushpattnaik:
# There are two changes to the documentation:

## 1. Environment Variable name mismatch between code and Vercel
The neon-managed vercel integration automatically sets the environment variables in Vercel like this :
image

So the build fails because it is not able to detect the `DIRECT_URL` variable and hence fails to find datasource.url 
brave_screenshot_vercel com (1)

Therefore it is safe to change it to `DATABASE_URL_UNPOOLED` since for other deployment platforms there would be no such auto env variable naming. And if there is, we might need to change from vercel part of the code to give correct variable names

## 2. Pre-requisite command for Vercel Build
Current command for build override (`npx prisma migrate deploy && npm run build`) fails because it is not able to detect the client from the generated schema folder because there is no schema generate without the` prisma generate `command.
brave_screenshot_vercel com

### DIFF
--- a/content/docs/guides/neon-managed-vercel-integration.md
+++ b/content/docs/guides/neon-managed-vercel-integration.md
@@ -156,7 +156,7 @@ This isolation allows you to test data and schema changes safely in each pull re
 1. Go to **Vercel Dashboard → Settings → Build and Deployment Settings**
 2. Enable **Override** and add your build commands, including migrations, for example:
    ```bash
-   npx prisma migrate deploy && npm run build
+   npx prisma generate && npx prisma migrate deploy && npm run build
    ```
 
 This ensures schema changes in your commits are applied to each preview deployment's database branch.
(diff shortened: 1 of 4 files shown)


danieltprice:
Reviewed and live-tested. Replicated the doc's Prisma 7 flow against a real Neon project: `prisma db push` connected via `DATABASE_URL_UNPOOLED` and synced the schema, and CRUD through the pooled `DATABASE_URL` + `@prisma/adapter-neon` worked end to end.

The rename `DIRECT_URL` → `DATABASE_URL_UNPOOLED` aligns these guides with the name Neon's console/Vercel integration actually provisions (and with the convention used across the rest of the docs), and the added `npx prisma generate` prevents a 'Prisma Client not generated' failure on preview builds. Also tidied the `Direct(Unpooled)` label spacing to `direct (unpooled)`. Thanks for the fix!

danieltprice (approved):
Live-tested end to end (db push via DATABASE_URL_UNPOOLED + CRUD via pooled adapter). Rename aligns with Neon's provisioned env var name; build fix is correct.

PR #4865, opened 2026-05-07
Fix stop value in example in the generate_series function tutorial

duffuniverse:
### Detail

Fix a few minor typos in the `generate_series` function tutorial


philip:
Thanks @duffuniverse for the fix, merged!

PR #4883, opened 2026-05-09
docs: update Neon Auth UI import paths

hjq041015:
## Problem

The current Neon Auth UI documentation imports styles and components directly from `@neondatabase/auth-ui`:

```css
@import "@neondatabase/auth-ui/tailwind";
```

```ts
import { AuthView } from "@neondatabase/auth-ui";
```

When using pnpm, this can fail with:

```txt
Can't resolve '@neondatabase/auth-ui/tailwind'
```

## Cause

`@neondatabase/auth-ui` is installed as a transitive dependency of `@neondatabase/neon-js` / `@neondatabase/auth`, so it is not always resolvable directly from application source code under pnpm's dependency isolation.

## Fix

Use the public exports from `@neondatabase/neon-js` instead:

```css
@import "@neondatabase/neon-js/ui/tailwind";
```

```ts
import { AuthView } from "@neondatabase/neon-js/auth/react/ui";
```

## Verification

I reproduced the issue in a Vite + React + Tailwind v4 + pnpm project.

After changing the imports above, the project builds successfully with:

```bash
pnpm build
```

### DIFF
--- a/content/docs/auth/quick-start/tanstack-router.md
+++ b/content/docs/auth/quick-start/tanstack-router.md
@@ -102,7 +102,7 @@ See [UI Component Styles](/docs/auth/reference/ui-components#styling) for altern
 <TwoColumnLayout.Block label="Add to src/styles.css">
 
 ```css
-@import '@neondatabase/auth-ui/tailwind';
+@import '@neondatabase/neno-js/ui/tailwind';
 ```
 
 </TwoColumnLayout.Block>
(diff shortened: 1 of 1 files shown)


thekauer:
Hi @hjq041015, thanks for the clean repro, the pnpm issue is real. You're right that the transitive `@neondatabase/auth-ui` doesn't resolve under pnpm's strict isolation.

We're going the other direction though. In [neon-js#111](https://github.com/neondatabase/neon-js/pull/111) we deprecated the `@neondatabase/neon-js/...` UI re-exports (they'll be removed in the next major) and are pushing users to install `@neondatabase/auth-ui` directly instead. That also fixes the pnpm resolution.

Codemod for existing imports:

```bash
npx -p @neondatabase/auth neon-auth-codemod --write <path>
```

#4895 updates the docs along those lines. Closing this in favour of that, thanks again for filing it!

PR #4896, opened 2026-05-11
fix: case studies filter tabs no longer break back button

rahulhgdev:
## What's the problem?

Clicking category filters on the case studies page was adding a new  browser history entry each time. So if you clicked through a few  filters and opened an article, the back button would take you through every filter you'd clicked instead of just going back to the page.

https://github.com/user-attachments/assets/9a920640-ae2d-4499-9abd-b3b479b62759

### Steps to reproduce

1. Go to https://neon.com/case-studies
2. Click a few filters - Startups, Fast-moving teams, AI Agents
3. Open any article
4. Hit back

Before: `article → #ai-agents → #fast-moving-teams → #startups → /case-studies`  
After: `article → /case-studies`

## What's the fix?

When a filter is clicked, instead of letting the browser push a new history entry, we now replace the current one using `replaceState`. The URL still updates so deep-linking and sharing still work - the back button just behaves correctly now.

Also added a `popstate` listener alongside the existing `hashchange` one. Without it, hitting back from an article wouldn't sync the active filter with the URL since back navigation fires `popstate`, not `hashchange`.

## File changed

```
src/components/pages/case-studies/cards/cards.jsx
```

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -235,6 +235,62 @@ const defaultConfig = {
         destination: '/?ref=tbm-p',
         permanent: true,
       },
+      // Branching page redirects — old slugs deleted in PR #4374 (Jan 2026)
+      {
+        source: '/branching/identifying-use-case',
+        destination: '/branching/introduction',
+        permanent: true,
+      },
+      {
+        source: '/branching/production-on-neon',
+        destination: '/branching/production-staging-workflows',
+        permanent: true,
+      },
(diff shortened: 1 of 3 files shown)


philip:
Thanks @rahulhgdev for the fix!

(also, sorry, Claude and I caused chaos in your PR; AGI isn't here yet...)

PR #4922, opened 2026-05-15
Fix view name in the PostgreSQL Drop View docs

duffuniverse:
### Detail

Fix view name in the PostgreSQL Drop View docs


danieltprice (approved):
Thank you @duffuniverse

PR #4940, opened 2026-05-20
docs: add steps for moving data from dev to prod branch

dhanushreddy291:
https://neon-next-git-fork-dhanushreddy291-update-d-e2fa60-neondatabase.vercel.app/docs/guides/move-branch-changes-to-production

### DIFF
--- a/content/docs/guides/move-branch-changes-to-production.md
+++ b/content/docs/guides/move-branch-changes-to-production.md
@@ -1,0 +1,163 @@
+---
+title: Move changes from a branch to production
+subtitle: Options for schema or data changes already made on a development branch
+summary: >-
+  Explains what to do when schema changes or data already exist on a
+  development branch and need to be moved back to production.
+enableTableOfContents: true
+updatedOn: '2026-05-20T09:47:02.219Z'
+---
+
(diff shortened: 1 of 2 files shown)


content/docs/guides/move-branch-changes-to-production.md
```diff
@@ -0,0 +1,163 @@
...
+- Lookup tables
+- Country or region data
+- Configuration records
+- Seed datasets
+
+Typical workflow:
+
+1. Export selected tables from the development branch
+2. Review or modify the exported data
+3. Import the data into production with a conflict strategy, such as `INSERT ... ON CONFLICT`
```
philip: Wondering if this should reference `--inserts --on-conflict-do-nothing` or how to clarify the upsert advice since it's near the `pg_dump` example that doesn't allow generating it.

dhanushreddy291: thanks @philip, have updated it

Previously, the main focus was on non-collision cases / smaller datasets, so I thought this check wasn’t necessary and only a simple note sentence would suffice.

Since adding it doesn’t affect the existing SQL code, I’ve included it now.

PR #4958, opened 2026-05-26
Fix example in the PostgreSQL CREATE TRIGGER Statement docs

duffuniverse:
### Detail

Add missing semicolon after `CREATE OR REPLACE FUNCTION` statement


philip:
Thanks again @duffuniverse !

PR #4979, opened 2026-05-28
fix: copy correct SVG source from navbar logo context menu

mezotv:
Static `.svg` imports resolve to a `StaticImageData` object, not a URL string, so the logo context-menu "Copy logo as SVG" action now fetches `logoToUse.src` instead of the object.

If you right click the logo on the landing page today and copy the svg it copies something completely unrelated hence this fix!


mezotv:
@danieltprice should be good now!

ruf-io (approved):

PR #5008, opened 2026-06-04
docs: fix anonymous role description and add SDK example

philip:
Issue spotted by user, code provided by Ruslan.

### DIFF
--- a/content/docs/data-api/access-control.md
+++ b/content/docs/data-api/access-control.md
@@ -7,7 +7,7 @@ summary: >-
   uses PostgreSQL's security model to enforce role privileges and Row-Level
   Security for database access control.
 enableTableOfContents: true
-updatedOn: '2026-06-04T01:09:28.144Z'
+updatedOn: '2026-06-04T14:48:01.456Z'
 ---
 
 <FeatureBetaProps feature_name="Neon Data API" />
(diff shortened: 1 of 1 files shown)


content/docs/data-api/access-control.md
```diff
@@ -42,12 +42,30 @@ When a client sends a valid Bearer token, the API switches to the `authenticated
 
 ### 2. The `anonymous` role
 
-**Used for:** Requests without a token.
-If a request arrives with no `Authorization` header, the API switches to the `anonymous` role.
+**Used for:** Requests from unauthenticated users.
+
+Anonymous access still uses a JWT, but your auth provider mints one automatically without requiring a user to sign in. Set `allowAnonymous: true` in the client config. The SDK fetches a short-lived anonymous token (`GET /token/anonymous`) on the first request, caches it, and sends it as `Authorization: Bearer <jwt>` on every query.
```
ruslantalpa: this code/statement is true only for the cases when people use neon-auth.
If they do not use neon-auth with data api (the used a 3rd party jwks url when configureing data-api) then it's up to them to figure out how get that auth provider to issue a token for the anonymous user.

ruslantalpa (approved):

PR #5021, opened 2026-06-05
Fix a few minor mistakes in the PL/pgSQL Exit Statement docs

duffuniverse:
### Detail

- Add missing semicolon in code examples.


ruf-io (approved):
Correct: each `do $$ ... $$` block should be terminated with a semicolon. Verified all three occurrences. Thanks!

PR #5025, opened 2026-06-07
Fix a few mistakes in the "PL/pgSQL Function Parameter Modes: IN, OUT, INOUT" docs

duffuniverse:
### Detail

- Add missing semicolon in code examples.
- Fix `film ID` in a query description.


ruf-io (approved):
Both fixes are correct: `end;$$;` properly terminates the CREATE FUNCTION statement (the `language plpgsql` clause precedes `as $$`), and the example calls `find_film_by_id(1)`, so "id 1" matches. Thanks!

PR #5026, opened 2026-06-07
Fix interval tutorial examples and terminology

rahulrao0209:
Updated examples and corrected terminology in the interval tutorial.


ruf-io (approved):
Verified all four fixes against the file: the SQL uses `INTERVAL '1 year 3 hours 20 minutes'`, the table column is `duration INTERVAL`, the subject "two functions" takes "allow", and on Postgres 14+ `EXTRACT()` labels the output column `extract`. Approving.

Minor optional follow-up (not blocking): the line just above still says EXTRACT "returns a value of type double precision", which is `numeric` on PG 14+. Could be tidied in a future pass. Thanks for the cleanup!

PR #5028, opened 2026-06-08
Fix record variable name in the PL/pgSQL Record Types docs

duffuniverse:
### Detail

- Fix record variable name


ruf-io (approved):
Correct: the loop variable is declared and used as `rec` (`rec.title` / `rec.length`), so the prose should say `rec`, not `r`. Thanks for the fix!

PR #5035, opened 2026-06-09
feat: docs-fact-conflicts workflow (cross-page fact conflict checker)

ruf-io:
## What

Adds a developer tool (not a site/runtime change) under `.claude/workflows/` that checks a single `content/docs` page for **facts that contradict the rest of the docs** — e.g. one page saying "5,000 branches/project" while another says "1,000".

It is a dynamic Claude Code workflow run via the `Workflow` tool with the target page as `args`. Nothing here is imported by the Next.js build.

## How it works

1. **Extract** — one agent pulls cross-referenceable facts (plan limits, prices, units, defaults, version numbers, product-name casing) from the target page, each tagged with a `qualifier` (which plan, per-branch vs per-project, legacy, unit).
2. **Search** — one batched agent finds other mentions of each fact across `content/docs` (grep, plus optional local semantic search).
3. **Judge** — per fact, a scope-aware agent decides whether a candidate genuinely conflicts (a different plan/tier, a legacy page, or a different unit is **not** a conflict).
4. **Verify** — an adversarial pass tries to refute each flagged conflict before it is reported, keeping false positives near zero.

## Two search modes

- **Default (grep-only, no deps):** `args: "content/docs/introduction/plans.md"`
- **Grep + local semantic search:** `args: { path: "...", semantic: true }` — a SQLite + `sqlite-vec` index of `content/docs`, embedded with `Xenova/all-MiniLM-L6-v2`, running **fully offline (no API key)**. Raises recall on paraphrased facts grep misses. Index build is incremental (sha1 per file), which also refreshes the page being checked.

Semantic deps live in `.claude/workflows/lib/package.json`, isolated from the website build; `lib/node_modules` and `lib/.cache` are gitignored.

## Validation

Run against `content/docs/introduction/plans.md`:

- Batched search cut subagents ~107 → ~50 and tokens ~3.4M → ~1.4M, same wall-clock.
- Semantic index: 577 files / 6,950 chunks; build + query verified running cleanly inside the workflow.
- Surfaced real, hand-confirmed conflicts, including `plans.md` "5,000 branches/project max" vs `ai-agent-integration.md` "1,000", and `plans.md` HIPAA "additional charge" vs `hipaa.md` "no additional cost".

## Notes / limitations (documented in the script header)

- The extract stage is non-deterministic — different runs surface different facts, so re-running improves coverage.
- Single-page mode does not catch intra-page self-contradictions or conflicts between two pages that both exclude the target; a full-corpus sweep would, and is left as a follow-up.
- Findings are leads to confirm by hand, not auto-verdicts.

### DIFF
--- a/.claude/commands/fact-check.md
+++ b/.claude/commands/fact-check.md
@@ -25,27 +25,56 @@ current `main` first. Let the user decide whether to proceed or update.
 
 ## Pick the engine (auto-detect)
 
+**First, resolve the target to an ABSOLUTE path** inside the worktree you mean to check
+(e.g. `/abs/path/<worktree>/content/docs/introduction/plans.md`). Both engines otherwise search
+`content/docs` relative to the session's cwd — silently checking whatever checkout the session sits
+in, which is the wrong tree when you mean to vet a different branch. Confirm which branch that
+worktree is on (warn if it is stale vs `main`), then pass absolute paths so the corpus root is
+unambiguous.
+
 Check whether the **`Workflow` tool** is available this session (e.g. `ToolSearch` for `Workflow`).
 
-- **Workflow available + single-page** → run the headless engine:
+- **Workflow available + single-page** → run the headless engine, passing the ABSOLUTE target so it
+  roots its search at the right worktree (the script derives `docsRoot` from an absolute target, or
+  pass `docsRoot` explicitly):
   ```
-  Workflow({ scriptPath: ".claude/workflows/docs-fact-conflicts.js", args: "<page path>" })
+  Workflow({ scriptPath: "<abs>/.claude/workflows/docs-fact-conflicts.js",
+             args: { path: "<abs>/content/docs/.../page.md" } })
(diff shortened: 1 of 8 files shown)


bgrenon:
Added a commit here that does two things:

1.   An agent-based fallback so the checker still runs when the Workflow tool isn't available. A Claude session drives the same extract → search → judge → verify pipeline directly, and a shared script merges the results so the output matches the Workflow version.
2. A full-corpus sweep mode. The current checker compares a single page against the rest; this adds a whole-docs pass that pulls facts from every page in a set and cross-checks them against each other — so it can catch conflicts between two pages that wouldn't obviously get compared.

Three new files, all additive (docs-fact-conflicts.js is unchanged):
- a shared synthesis + coverage script used by both paths
- a runbook for the agent orchestration (single-page and full-corpus)
- a /fact-check command that auto-detects the Workflow engine and falls back to the runbook

PR #5043, opened 2026-06-10
fix: remove backtick formatting from linked inline code

danieltprice:
Links wrapped in backticks render with code styling that hides the link indicator. Convert all [` code `](url) patterns to plain [code](url) links across docs and guides.


danieltprice:
we're doing a style revision instead

PR #5061, opened 2026-06-12
chore: promote Lakebase Search in topbar; remove undeployed Cmd+K changelog entry

ruf-io:
## What

- **Topbar banner** (`content/config/topbar.yaml`): updated to promote **Lakebase Search** (scalable vector + BM25 full-text search) from the 2026-06-12 changelog, linking to the changelog entry.
- **Changelog** (`content/changelog/2026-06-12.md`): removed the `Cmd+K support for the Neon Console` section. The command bar feature was not fully deployed, so it shouldn't be announced yet.
- Updated the changelog title from "Lakebase Search, command bar, and more" to "Lakebase Search, CLI updates, and more".
- Deleted the now-orphaned `public/docs/changelog/command_k.png` image.

## Why

The Cmd+K command bar was never fully deployed, so it needs to come out of the published changelog. The topbar instead highlights Lakebase Search, the marquee item in the same changelog.


content/config/topbar.yaml
```diff
@@ -1,4 +1,4 @@
-text: 'All paid plans now include 5x more free data transfer: 500 GB per month, up from 100 GB. Read the details here.'
+text: 'New in Neon: Lakebase Search brings scalable vector and BM25 full-text search to a single Postgres database. See what shipped this week.'
```
ruf-io:
```suggestion
text: 'New in Neon: Vector and BM25 full-text search extensions built for lakebase architecture, plus major CLI updates. See what shipped this week.'
```

PR #5179, opened 2026-06-30
Fix variable name in the PostgreSQL EXP() Function docs

duffuniverse:
## Detail

- Fix a variable name.


danieltprice (approved):
Thank you @duffuniverse

PR #5185, opened 2026-07-01
Fix code example description in the PostgreSQL MOD() Function docs

duffuniverse:
### Detail

- Fix a code example description


danieltprice (approved):
Thank you @duffuniverse

PR #5195, opened 2026-07-02
simplify lakebase search enable flow for default preload

bgrenon:
## What this does

  Neon now preloads `lakebase_vector` and `lakebase_text` by default (neon-cloud#8195), so these docs should lead with
  `CREATE EXTENSION` instead of a manual preload + restart.

  - **`lakebase-search-get-started.md`** — collapses the three enable steps (Enable preload libraries → Restart → Enable
  extensions) into one **Enable the extensions** step. Leads with `CREATE EXTENSION`; the API preload + restart path is
  kept as a `<details>` fallback for computes created before the default.
  - **`lakebase-search.md`** (hub) — "Enable it once, then install…" → "preloaded by default, install… right away."
  - **`lakebase-text.md` / `lakebase-vector.md`** — "must be enabled first" → "preloaded by default, install directly."
  - **`pg-extensions.md`** — lakebase rows drop the "Requires enablement" note to match every other default-preloaded
  extension (no preload note); the shared-preload fallback in other pages points here as the canonical reference.
  - **`migrate-pg-search-to-lakebase-text.md`** — enable step leads with `CREATE EXTENSION`, manual path as an error
  fallback pointing to `pg-extensions`.

  ## ⚠️ Hold until default preload is live in production

  Every change asserts "preloaded by default." That's true in code (merged neon-cloud#8195, Jul 1) but is
  **compute-version gated** and not yet rolled out to prod computes. **Do not merge until verified live** — otherwise
  users hit `ERROR: ... must be loaded via shared_preload_libraries`.

  Verify: on a fresh Neon project, `available_preload_libraries` shows `is_default: true` for both libs and `CREATE
  EXTENSION lakebase_text` works with no manual preload.


content/docs/ai/lakebase-search-get-started.md
```diff
@@ -14,43 +14,35 @@ This guide sets up Lakebase Search on a Neon project: enabling both extensions,
...
 - Node.js 18 or later
 - An [OpenAI API key](https://platform.openai.com/api-keys) for generating embeddings
 
 <Steps>
 
-## Enable the preload libraries
+## Enable the extensions
 
-`lakebase_vector` and `lakebase_text` rely on preloaded libraries that aren't enabled by default. Check your project's current state in the [Neon SQL Editor](/docs/get-started/query-with-neon-sql-editor):
+`lakebase_vector` and `lakebase_text` are preloaded by default on Neon (Postgres 16 and later), so you can install them directly. Run the following in the [Neon SQL Editor](/docs/get-started/query-with-neon-sql-editor) or any connected Postgres client:
```
philip: This framing ("X is preloaded by default, so you can install X") is confusing now. Also, other pg extensions with shared libraries (e.g., timescaledb) don't reference them at all as it's an implementation detail.

Two potential options come to mind: (a) don't mention preloaded libraries in lakebase search docs at all, or (b) mention it in one isolated place (per ext). The one consideration that comes to mind is if a user modifies their projects' shared libraries and forgets to add them, then they'll be issues, but that's uncommon. Existing users who already modified their project's shared libraries and didn't add them would have issues, so I vote for (b).

My comment applies to this entire PR, I won't comment on every reference. Where (b) applies may mean on each extension's page and not mentioning at all in getting started.

That said, the general shared library documentation also requires an update to include lakebase_*.

content/docs/extensions/pg-extensions.md
```diff
@@ -46,9 +46,9 @@ Neon supports the Postgres extensions shown below. Extension versions may differ
...
-| [lakebase_text](/docs/extensions/lakebase-text)                                                  |       - |       - | 0.1.0-dev | 0.1.0-dev | 0.1.0-dev | Requires enablement. PG16+. See [Lakebase Search](/docs/ai/lakebase-search).                                                                                                                                                                                                                                                                                                                                                                                  |
+| [lakebase_text](/docs/extensions/lakebase-text)                                                  |       - |       - | 0.1.0-dev | 0.1.0-dev | 0.1.0-dev | PG16+. See [Lakebase Search](/docs/ai/lakebase-search).                                                                                                                                                                                                                                                                                                                                                                            |
```
philip: The columns indicate avail version, remove `PG16+`?

content/docs/ai/lakebase-search-get-started.md
```diff
@@ -14,43 +14,35 @@ This guide sets up Lakebase Search on a Neon project: enabling both extensions,
 
 ## Prerequisites
 
-- A Neon project. You enable Lakebase Search on it in the first step below.
+- A Neon project
 - Postgres 16 or later (Lakebase Search requires PG16+)
 - Node.js 18 or later
 - An [OpenAI API key](https://platform.openai.com/api-keys) for generating embeddings
```
philip: Consider clarifying that this is not required, in that it's specific to how this particular getting started guide chose to do the vector embeddings.

philip (commented):
Left a few comments to ponder for this interesting topic :)

bgrenon:
moved to: https://github.com/neondatabase/website/pull/5585

PR #5203, opened 2026-07-03
feat(ai-gateway): add /models.json catalog endpoint

andrelandgraf:
## Summary

Adds a JSON endpoint at **`/models.json`** that describes every model served by the **Neon AI Gateway** — intended as the **source of truth for the AI Gateway docs** (`content/docs/ai-gateway/models.md`).

The response **mirrors the [models.dev API](https://models.dev/api.json) shape** so the same parsing code works against either: the top-level key is the provider id (`neon`), whose `models` map is keyed by the gateway's short catalog id. Each model carries its underlying `provider`, capability flags (`attachment`, `reasoning`, `tool_call`, `temperature`), `modalities`, `limit`, and `cost`.

```jsonc
{
  "neon": {
    "id": "neon",
    "name": "Neon",
    "npm": "@ai-sdk/openai-compatible",
    "api": "${NEON_AI_GATEWAY_BASE_URL}/ai-gateway/mlflow/v1",
    "env": ["NEON_AI_GATEWAY_BASE_URL", "NEON_AI_GATEWAY_TOKEN"],
    "doc": "https://neon.com/docs/ai-gateway/models",
    "models": {
      "claude-sonnet-4-6": {
        "id": "claude-sonnet-4-6",
        "provider": "anthropic",
        "attachment": true, "reasoning": true, "tool_call": true, "temperature": true,
        "modalities": { "input": ["text", "image", "pdf"], "output": ["text"] },
        "limit": { "context": 1000000, "output": 64000 },
        "cost": { "input": 3, "output": 15, "cache_read": 0.3, "cache_write": 3.75 }
      }
      // …35 more
    }
  }
}
```

## What's included

- **36 models** across Anthropic, OpenAI, Google, Meta, and Alibaba — the full set the gateway serves today.
- Per-model `provider` (the one extension beyond models.dev's schema, since the docs table needs it).
- Follows the existing repo convention (`.well-known/ai-catalog.json/route.js`): a committed `data.json` served by a `force-static` route handler, so the endpoint is self-contained (no runtime dependency on models.dev).

## How the data was produced (accuracy)

Every entry is **live-verified against a real Neon AI Gateway branch** (`us-east-2`):
- **text output** — minimal completion returned 200 for all 36.
- **image input (`attachment`)** — sent an image part; 200 ⇒ supported. This corrects a few models whose canonical (multimodal) metadata disagrees with what Neon serves — e.g. `qwen35-122b-a10b` and `meta-llama-3-3-70b-instruct` are **text-only** on the gateway.
- **tool calling (`tool_call`)** — sent a function tool; **all 36 support it**.

The `data.json` is generated by resolving the Neon provider definition in models.dev (a companion PR updates that provider: anomalyco/models.dev#3019) and stripping generation-internal fields. Open-weight pricing uses the Databricks pay-per-token rate (DBU × \$0.070/DBU); proprietary models use provider list price.

## Verification
- `GET /models.json` → `200`, `application/json`, 36 models (verified locally via `next dev`).
- `eslint` clean on the new route.

## Notes
- Not served by the gateway and intentionally excluded: `gpt-5-5` (gated), `claude-sonnet-5`, `claude-fable-5`, `meta-llama-3-1-405b-instruct`, the Gemini image-output models, and embeddings (no `/embeddings` route).
- `data.json` is a generated artifact; when the catalog changes it can be regenerated from the models.dev Neon provider. Happy to add a `scripts/` generator + `check:*` guard if you'd like it wired into CI.

### DIFF
--- a/src/app/models.json/data.json
+++ b/src/app/models.json/data.json
@@ -1,0 +1,1589 @@
+{
+  "neon": {
+    "id": "neon",
+    "name": "Neon",
+    "npm": "@ai-sdk/openai-compatible",
+    "api": "${NEON_AI_GATEWAY_BASE_URL}/ai-gateway/mlflow/v1",
+    "env": [
+      "NEON_AI_GATEWAY_BASE_URL",
+      "NEON_AI_GATEWAY_TOKEN"
+    ],
+    "doc": "https://neon.com/docs/ai-gateway/models",
+    "models": {
+      "claude-haiku-4-5": {
+        "id": "claude-haiku-4-5",
+        "name": "Claude Haiku 4.5 (latest)",
+        "provider": "anthropic",
+        "family": "claude-haiku",
+        "attachment": true,
+        "reasoning": true,
+        "reasoning_options": [
+          {
+            "type": "toggle"
+          },
+          {
+            "type": "budget_tokens",
+            "min": 1024,
+            "max": 63999
+          }
+        ],
+        "tool_call": true,
+        "temperature": true,
+        "open_weights": false,
+        "knowledge": "2025-02-28",
+        "release_date": "2025-10-15",
+        "last_updated": "2025-10-15",
+        "modalities": {
+          "input": [
+            "text",
+            "image",
+            "pdf"
+          ],
+          "output": [
+            "text"
+          ]
+        },
+        "limit": {
+          "context": 200000,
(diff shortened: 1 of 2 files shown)


philip (approved):
:shipit:

PR #5204, opened 2026-07-03
ci(ai-gateway): keep /models.json in sync with models.dev

andrelandgraf:
## Summary

Follow-up to #5203. Adds a **drift check** that keeps the Neon AI Gateway catalog served at **`/models.json`** in sync with its upstream source of truth — the `neon` provider in the [models.dev API](https://models.dev/api.json), which Neon maintains.

- **`.github/workflows/models-sync-check.yml`** — scheduled (daily) + `workflow_dispatch` job that runs the check and **fails on any drift**.
- **`src/scripts/check-models-sync.js`** — fetches `https://models.dev/api.json`, extracts the `neon` provider, and compares it to the committed `src/app/models.json/data.json` (what backs `/models.json`). It fails on:
  - models present in one catalog but not the other, and
  - per-model capability drift (`name`, `family`, `attachment`, `reasoning`, `reasoning_options`, `tool_call`, `temperature`, `structured_output`, `open_weights`, `knowledge`, dates, `modalities`, `limit`, `cost`, `status`).
  - Endpoint-specific fields are ignored on both sides so they don't cause false positives: the `provider` convenience field we add, and `description`/`benchmarks`/`weights`/`links`.
- **`package.json`** — `npm run check:models-sync` for local runs (`--ci` terse, `--json` machine-readable; `NEON_MODELS_URL=https://neon.com/models.json` compares the live endpoint instead of the repo file).
- Also **regenerates `data.json`** to pick up the latest models.dev Neon changes (adds `qwen35-122b-a10b`'s `reasoning_options`).

## Heads up — currently reports drift (expected)

The companion models.dev PR ([anomalyco/models.dev#3019](https://github.com/anomalyco/models.dev/pull/3019)) that adds these 12 models and removes `gpt-5-5` is **merged but not yet deployed to `models.dev/api.json`**. Until it redeploys, the check reports the expected delta:

```
In models.dev but missing from /models.json (1):  gpt-5-5
In /models.json but not in models.dev (12):        + gpt-5-3-codex, claude-opus-4-8, gemma-3-12b, ... 
Field drift (4 models): gemini-2-5-pro, gemini-3-1-pro, gemini-3-pro, gpt-5-4 (cost)
```

Once models.dev redeploys the merged `neon` provider, the daily job goes green (both catalogs derive from the same TOMLs). The workflow is **schedule + manual only** (not a PR gate), so it won't block merges.

## Test plan
- [x] `node src/scripts/check-models-sync.js` exits `1` on drift, `0` when in sync; `--json` and `--ci` modes verified.
- [x] `eslint` clean on the new script.

### DIFF
--- a/.github/workflows/models-sync-check.yml
+++ b/.github/workflows/models-sync-check.yml
@@ -1,0 +1,37 @@
+name: AI Gateway Models Sync
+
+# Verifies the Neon AI Gateway catalog served at /models.json stays in sync
+# with the upstream models.dev `neon` provider (https://models.dev/api.json).
+# Fails on any catalog or capability drift.
+
+on:
+  schedule:
+    # Every day at 8:00 AM UTC
+    - cron: '0 8 * * *'
+  workflow_dispatch: # Allows manual trigger from the GitHub UI
+
+permissions:
+  contents: read
+
+jobs:
+  models-sync:
+    name: 'Check /models.json vs models.dev'
+    runs-on:
+      group: neondatabase-protected-runner-group
+      labels: linux-ubuntu-latest
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@4d991eb9b905ef189e4c376166672c3f2f230481 # v2.11.0
+        with:
+          egress-policy: audit
+
+      - name: Checkout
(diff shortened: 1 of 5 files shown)


philip (approved):
Nice! 🛳️

PR #5214, opened 2026-07-07
docs(config-as-code): bring neon.ts docs up to parity with last 30 days of code + IA audit

bgrenon:
## Summary
Adds the missing `@neon/config-runtime` programmatic-API reference and brings `neon.ts` docs up to parity with the last 30 days of `neon-pkgs` changes, plus fixes from cross-review and a persona-driven IA read-through.

## Changes
- **`reference/config-runtime.md`** (new): documents the `@neon/config-runtime` programmatic API (`inspect`, `plan`, `apply`, `pullConfig`, `pushConfig`, injectable bundler) for CI/power-user use cases that call it directly instead of via the CLI
- **`reference/neon-ts.md`**: added `env.branch.name`/`NEON_BRANCH` (a real parity gap found outside the original findings list)
- **`cli/config.md`**: clarified `config init` scaffolding, added `--allow-protected` to a CI example, documented plan-then-apply safety and `PushResult`, clarified the CI auth-fallback path
- **`compute/functions/deploy.md`, `environment-variables.md`**: fixed `NEON_ESBUILD_PATH` guidance — it only applies to the CLI's bundler, not `config-runtime`'s (would silently mislead an agent copying it into a CI script)

## Fixed after cross-review (factual corrections against real source)
- `createBranch` calls the API directly — it does NOT run `neon checkout` internally (previously described backwards)
- `apiHost` is not supported by every function — `createBranch`'s options type doesn't have it
- `pushConfig`'s `branchExists: false` never creates a branch; it always requires an existing `branchId`


bgrenon:
IA-audit follow-up applied in 5217a49f5.

Fixed locally scoped docs findings:
- Expanded `neon config init` docs with generated `neon.ts` shape, import/export style, idempotency behavior, and non-interactive setup guidance.
- Added missing `--allow-protected` to the copy-paste `neon config apply` CI/agent example.
- Updated `@neon/config-runtime` CI sample to plan first and require an explicit `NEON_APPLY=true` gate before applying.
- Documented `PushResult` fields near `plan`/`apply`/`pushConfig`, including what `applied` represents.
- Clarified that CI must pass `apiKey` or set `NEON_API_KEY` instead of relying on local CLI credential fallback.

Validation: `npx markdownlint content/docs/cli/config.md content/docs/reference/config-runtime.md content/docs/reference/neon-ts.md`, `npm run check:docs`, and pre-push `npm run test:unit:run` all passed.

danieltprice (approved):

bgrenon:
Pulled the two `env.branch?.name // NEON_BRANCH` additions (one code sample + accompanying prose in `neon-ts.md`, one code sample in `compute/functions/environment-variables.md`) in commit 698774fef.

Those asserted `NEON_BRANCH` holds the branch **name**, but the Functions runtime (`neon-cloud`'s `branchenv.go`) injects the branch **ID**, not the name -- a real SDK-vs-runtime contradiction, not just wording. This overlaps with an already-open, unresolved ticket (LKB-14741) filed against the pre-existing `environment-variables.md` table row that has the same problem; didn't want this PR to add two *more* copies of the same unverified claim in two new places ahead of that ticket landing on a confirmed answer.

Everything else in this PR (`config-runtime.md`, `deploy.md`, `cli/config.md`, `navigation.yaml`) is unaffected and unrelated to this issue. Noted the overlap on LKB-14741 so the eventual fix covers all three locations (the original table row + these two) together.

bgrenon:
One more fix after a full source-verification pass on the rest of this PR's claims (~55+ across all 6 touched files): commit d98bd4ded corrects `config-runtime.md`'s `confirm` field description, which claimed unresolvable conflicts are specifically about immutable fields like **region or Postgres version**.

That's not accurate today: `@neon/config`'s schema has no field for declaring a desired region/Postgres version, the diff engine (`diff.ts`) can't currently produce that conflict kind, and `isImmutableConflict()` is hardcoded to `return false` -- the 'immutable, recreate the project' code path is unreachable dead code right now. Dropped the specific example while keeping the accurate part (some conflicts do bypass `confirm` and throw `PushConflictError` directly).

Everything else checked (all CLI flags/defaults, SDK function signatures, error codes, auth resolution order, nav wiring, every internal link/anchor) matched the real `neon-pkgs` source exactly, and nothing in this PR is feature-flag-gated.

PR #5218, opened 2026-07-08
Add /oneshot skill for testing agentic one-shot setup of Neon guides

danieltprice:
## Summary
- Adds a `/oneshot` skill that tests whether an AI coding agent can complete a Neon setup guide in a single shot — no clarifying questions, no leaving the doc page — the way a developer would when dropping the guide link into Cursor or Claude Code
- Provisions a real throwaway Neon project, runs an isolated test-subject agent (`.claude/agents/oneshot-subject.md`, restricted to no web access and no Neon MCP tools) against the doc content alone, and independently verifies the result against the live database rather than trusting the subject's self-report
- Produces a report per guide: pass/fail verdict, token cost, a 6-criterion rubric (self-containment, step ordering, placeholder clarity, code block completeness, currency, failure-mode guidance), and a friction log tied to specific doc headings with concrete rewrite suggestions
- First supported slice covers 5 guides (nextjs, python, django, go, express), chosen because all five already ship a companion `public/prompts/<slug>-prompt.md` file to compare against

Initial run against all 5 guides in the slice (all passed, independently verified) surfaced one clear cross-guide pattern: guides that explicitly say which driver to pick and why (express, go, django) score much better on self-containment than guides that just list options without a default (nextjs) — and failure-mode guidance is weak everywhere except django, which has a real troubleshooting section worth porting to the others.

## Test plan
- [x] Ran `/oneshot guides/nextjs`, `/oneshot guides/python`, `/oneshot guides/django`, `/oneshot guides/go`, `/oneshot guides/express` end to end — all passed, each independently verified via `run_sql` against a live throwaway Neon project
- [x] Confirmed no credentials, connection strings, or secrets appear in either committed file
- [x] Pre-push unit tests pass (294/294)


danieltprice:
Closing this. The /oneshot skill is being kept as a local, machine-global Claude skill rather than committed to the public repo, since its SKILL.md references internal infrastructure details (package-registry proxies, environment policy) that don't belong in a public repo. The doc improvements it surfaced were already merged separately (#5220). No public-repo artifact is needed for the tool itself.

PR #5226, opened 2026-07-09
Daniel/ai gateway models source of truth

danieltprice:
Adds `src/data/ai-gateway-models.yaml` as the source of truth for every model on Neon's AI Gateway, replacing the models.dev-derived generation that had drifted from what the gateway actually serves. Includes a `/update-ai-models` command to keep it current and a README explaining the schema and validation approach.


danieltprice:
Closing this one. The AI Gateway models source-of-truth work has been handled elsewhere, so this PR is no longer needed.

PR #5250, opened 2026-07-13
feat: agent-discovery fixes (/openapi.json, MCP card) + unified verifiers for discovery & docs↔API consistency

andrelandgraf:
## Summary

Two agent-discovery fixes that ship to prod, plus the verification that keeps them (and the docs) correct going forward.

**Prod-facing fixes** — the [integrations.sh scorecard for neon.com](https://integrations.sh/neon.com/) reports **3/8** conventions; two misses are trivial because the assets already exist:

- **Canonical OpenAPI path** — the Neon OpenAPI spec is published at `https://neon.com/api_spec/release/v2.json` (and linked from `/.well-known/api-catalog`), but discovery tooling probes `/openapi.json`. This aliases `/openapi.json` to the published spec (`application/json`).
- **MCP server card shape** — `/.well-known/mcp/server-card.json` used `endpoint` instead of the expected `url` and carried no `authentication` metadata. Added `url` + an OAuth `authentication` block mirroring `https://mcp.neon.tech/.well-known/oauth-authorization-server`; `endpoint` kept for back-compat.

Together these move the scorecard 3/8 → 5/8. No invented endpoints.

## Verification: two verifiers, split by concern

Rather than one mega-check, the verification is grouped by **concern + owner + failure semantics**. Both share the same design (spec-aware inventory / pure lib / offline-on-PR + strict-on-schedule / drift issue), but stay separate because a docs-pipeline change should never turn a *discovery* check red, and blocking vs advisory severity differs.

### 1. Agent-discovery surfaces — `agent-discovery-verify.yml`

*"Is Neon correctly and discoverably described to agents?"* Folded in from #5249: single source of truth (`src/constants/agent-discovery.js`) + spec-aware catalog (`config/agent-endpoints.yaml`) + unified validator (`scripts/verify-agent-endpoints.mjs`) covering skills, MCP card, API catalog, ai-catalog, models.json (replaces the old `models-sync-check`), llms.txt, and the `/openapi.json` alias. Offline on PRs; live daily with a drift issue.

### 2. Docs ↔ API consistency — `docs-api-consistency.yml`

*"Are the docs faithful to and complete against the API surface + `@neon/sdk`?"* Consolidates the previously separate `sdk-docs-check` + `sdk-docs-drift` into one checker, and **adds the coverage check requested** — every operation reflected across all three surfaces (OpenAPI spec ↔ docs ↔ `@neon/sdk` raw):

- **`scripts/check-docs-api-consistency.mjs`** (replaces `check-sdk-docs.mjs`):
  - **Snippet correctness (blocks PRs):** fenced `ts`/`js` `@neon/sdk` samples must resolve (`raw.*`, `neon.*`, imports).
  - **Operation coverage (advisory on PRs; fails weekly `--strict`):** documented ops ↔ `@neon/sdk` raw ↔ (in strict) the live spec's `operationId`s.
- **`scripts/lib/api-coverage.mjs`** — pure, unit-tested coverage diff (in `operationId` space via `toSdkMethodName`). This **replaces the old operation-set logic in `check-sdk-docs.mjs` that read the gitignored `src/data/api-ref`** and was therefore unreliable in CI.
- **`content/docs/api-operation-ids.json`** — new committed manifest of documented `operationId`s emitted by `generate-api-ref.mjs`, so coverage runs offline without regenerating gitignored data. Kept fresh by `check-api-ref-generated.mjs` (was `check-api-ref-nav.mjs`; now guards nav **and** the manifest) in `prebuild`.

### Renames (names now match behavior)

| Before | After |
|---|---|
| `scripts/check-sdk-docs.mjs` | `scripts/check-docs-api-consistency.mjs` |
| `check:sdk-docs` | `check:docs-api-consistency` |
| `scripts/check-api-ref-nav.mjs` | `scripts/check-api-ref-generated.mjs` |
| `check:api-ref-nav` | `check:api-ref-generated` |
| `.github/workflows/sdk-docs-{check,drift}.yml` | `.github/workflows/docs-api-consistency.yml` |

## Test plan

- [ ] After deploy, `curl -sSI https://neon.com/openapi.json` returns 200 + `content-type: application/json`
- [ ] `curl -sS https://neon.com/.well-known/mcp/server-card.json | jq '{url, authentication}'` shows the new fields
- [ ] Re-run "Map integration surface" on integrations.sh; OpenAPI + MCP card rows flip to ✓
- [x] `node scripts/verify-agent-endpoints.mjs` → 7/7 discovery endpoints pass
- [x] `npm run check:docs-api-consistency` (offline) and `--strict` (live spec) both pass; spec ↔ docs ↔ `@neon/sdk` aligned at **163/163/163**
- [x] `npx vitest run scripts/lib/api-coverage.test.js scripts/check-docs-api-consistency.test.js` → 20 pass
- [x] `npm run generate:api-ref && npm run check:api-ref-generated` → nav + manifest up to date


next.config.js
```diff
@@ -2628,6 +2628,14 @@ const defaultConfig = {
           source: '/api_spec/release/v2.json',
           destination: 'https://dfv3qgd2ykmrx.cloudfront.net/api_spec/release/v2.json',
         },
+        // Canonical OpenAPI path probed by agent-discovery tooling (e.g. integrations.sh).
+        // Aliases the published Neon API spec (the same document served at
+        // /api_spec/release/v2.json — the canonical public URL referenced by
+        // @neon/sdk and the docs) so /openapi.json resolves as application/json.
+        {
+          source: '/openapi.json',
+          destination: 'https://neon.com/api_spec/release/v2.json',
```
philip: This isn't working (404): https://neon-next-git-feat-integrations-sh-openapi-27ee32-neondatabase.vercel.app/openapi.json needs to be in beforeFiles (see catch-all comments there).  

Also a nit: this v2.json destination is itself a rewrite to https://dfv3qgd2ykmrx.cloudfront.net/api_spec/release/v2.json so we just point at that directly.

philip (approved):
Nice, love it! I left a comment to consider.

next.config.js
andrelandgraf: TIL! I actually told my model NOT to use the cloudfront link because it felt like not what I want :D Updating!

andrelandgraf: Fixed in 2b67eccaa. Both points addressed:

- Moved `/openapi.json` from `fallback` → `beforeFiles` (right after the index `.md` rewrites), with a comment explaining the same catch-all reasoning: as a single top-level segment it was being intercepted by the `[slug]` catch-all before `fallback` could fire.
- Pointed the destination straight at `https://dfv3qgd2ykmrx.cloudfront.net/api_spec/release/v2.json` (what `/api_spec/release/v2.json` already targets) instead of the neon.com canonical URL — one hop, and rewrites don't chain so a relative `/api_spec/...` destination wouldn't resolve anyway.

Smoke tested locally against `next dev`:
- `GET /openapi.json` → `200 application/json`, serving the real spec (`openapi: 3.0.3`, title "Neon API", 111 paths)
- `GET /api_spec/release/v2.json` → `200` (unchanged)

Will re-run `curl -sSI https://neon.com/openapi.json` after this deploys to confirm on prod.

PR #5238, opened 2026-07-10
fix: correct stale links, fabricated command, and outdated claims in ai-skills-corpus

bgrenon:
## Summary

Four mechanical fixes to the `public/docs/ai/skills/` corpus:

1. **Stale doc links (15 occurrences across 7 files):** links pointed at pre-restructure paths (`reference/cli-*.md`, `reference/neon-cli.md`) that no longer exist. Updated to the current `cli/*.md` and `cli.md` locations in:
   - `neon/SKILL.md`
   - `neon-postgres/SKILL.md`
   - `neon-postgres/references/devtools.md`
   - `neon-postgres/references/getting-started.md`
   - `neon-postgres/references/neon-cli.md`
   - `neon-postgres-branches/SKILL.md`
   - `neon-postgres-agent-platforms/scripts/auth-users.ts`

2. **Fabricated `neon sql` command** in `neon-postgres/references/neon-cli.md`: the CLI has no `neon sql` command. Replaced the two invented examples with the real `neon psql <branch> -- -c "..."` / `neon psql <branch> -- -f <file>` forms, verified against `content/docs/cli/psql.md`.

3. **Wrong VS Code extension id** in `neon-postgres/references/devtools.md`: `neon.neon-vscode` doesn't exist. Replaced with the real `databricks.neon-local-connect` id and corrected the marketplace search/publisher text ("Neon - Serverless Postgres" by Databricks), matching what `neon/SKILL.md` and `neon-postgres/SKILL.md` already say elsewhere in the corpus.

4. **Stale "planned sunset" claim** about the `GET /consumption_history/account` endpoint in `neon-postgres-agent-platforms/SKILL.md` (2 locations) and `references/pricing-and-plan-features.md` (1 location): this endpoint has already been removed from the API (see changelog 2026-06-19), so "deprecated with planned sunset of 2026-06-01" is now false. Replaced with a note that the legacy endpoint has been retired, use the v2 per-project endpoint instead.

No other files or content were touched.

## Test plan

- [x] Verified each stale link target exists in `content/docs/cli/` before rewriting
- [x] Verified real `neon psql` syntax against `content/docs/cli/psql.md`
- [x] Verified extension id/name against `content/docs/local/vscode-extension.md` and `content/docs/cli/init.md`
- [x] Verified the `/consumption_history/account` endpoint was actually removed (changelog 2026-06-19) before removing the sunset claim
[...]

PR #5266, opened 2026-07-15
fix: correct @neon/env import subpath and align auth API ref link in ai-skills

andrelandgraf:
## Summary

Two fact-checked corrections to the hosted AI skills corpus (`public/docs/ai/skills`), found while reconciling the hosted skills against the source repos (`neondatabase/agent-skills`, `neondatabase/neon-for-agent-platforms`).

- **`neon/SKILL.md`: `@neon/env/v1` → `@neon/env`** (2 occurrences). The `/v1` subpath does **not** exist in `@neon/env`'s package `exports` map — only `.` is exported (verified in `neondatabase/neon-pkgs/packages/env/package.json` and its README, which imports `from "@neon/env"`). Only `@neon/config` is versioned with `/v1`, so the convention was over-applied here. The documented import as written would fail to resolve.
- **`neon-postgres-agent-platforms/scripts/auth-users.ts`: auth API reference link.** Point the create-user reference at `https://neon.com/docs/reference/api/auth/create-branch-neon-auth-new-user` (canonical neon.com docs path) instead of `https://api-docs.neon.tech/reference/createbranchneonauthnewuser`. Both URLs currently return 200, but this aligns the hosted copy with the **later-updated** source of truth in `neondatabase/neon-for-agent-platforms` (PR #17, merged ~18 min after the hosted corpus fix PR #5238).

## Fact-check notes

- `@neon/env` has no `/v1` export → `/v1` import is broken. Fact-check overrides recency here.
- Both auth-ref hosts return real 404s on bad paths, so both are valid; tie broken by "later change wins" → source repo's neon.com path.

## Test plan

- [ ] Confirm `@neon/env` (no `/v1`) is the intended public import.
- [ ] Confirm the neon.com auth API reference path is preferred over `api-docs.neon.tech`.

### DIFF
--- a/public/docs/ai/skills/neon-postgres-agent-platforms/scripts/auth-users.ts
+++ b/public/docs/ai/skills/neon-postgres-agent-platforms/scripts/auth-users.ts
@@ -45,7 +45,7 @@ function printMeta(): void {
           "https://neon.com/docs/auth/guides/manage-auth-api",
           "https://neon.com/docs/auth/guides/user-management",
           "https://neon.com/docs/manage/roles",
-          "https://api-docs.neon.tech/reference/createbranchneonauthnewuser",
+          "https://neon.com/docs/reference/api/auth/create-branch-neon-auth-new-user",
         ],
         envForThisScript: {
           create: [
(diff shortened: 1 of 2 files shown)


philip (approved):
:shipit:  

(minor note: PR description meant to ref [#17](https://github.com/neondatabase/neon-for-agent-platforms/pull/17))

PR #5303, opened 2026-07-16
Fix syntax of the PostgreSQL FIRST_VALUE() function in the docs

duffuniverse:
### Detail 

- Remove a duplicate.


danieltprice (approved):
Thank you @duffuniverse

PR #5323, opened 2026-07-20
docs(ai-gateway): correct TPM error text and gateway env var count

bgrenon:
Doc-only fixes from a docs-vs-product audit. Two independent error-text/wording corrections across 3 files, 1 line each. **Not auto-merged** — awaiting your review.

### 1. Rate-limit error message text (`models.md`, `troubleshooting.md`)
The docs quoted an error string the AI Gateway never actually returns, so anyone troubleshooting from the docs would be searching for a message that doesn't exist.

- **Before:** `ai gateway TPM limit exceeded for model "<model-id>"`
- **After:** `ai gateway per-minute token limit exceeded for model "<model-id>"`
- **Why the new text is right:** "per-minute token limit exceeded" is the message the gateway actually returns when you exceed the token rate limit; the old "TPM limit exceeded" wording is never emitted.

### 2. Gateway env var count (`authentication.md`)
The credential snippet no longer ships four env vars.

- **Before:** "The snippet includes **all four** gateway env vars"
- **After:** "The snippet includes **both** gateway env vars"
- **Why the new text is right:** the snippet now exposes exactly two variables — `NEON_AI_GATEWAY_TOKEN` and `NEON_AI_GATEWAY_BASE_URL`; the two older OpenAI-alias variables that made it "four" were removed. "Two" also matches this same page's own env-var table further down (the page currently contradicts itself).

## Files
| File | Change |
|------|--------|
| `content/docs/ai-gateway/models.md` | error string → "per-minute token limit exceeded" |
| `content/docs/ai-gateway/troubleshooting.md` | same error string, second occurrence |
| `content/docs/ai-gateway/authentication.md` | "all four" → "both" env vars |

## Live test results

We exercised these AI Gateway claims against a live throwaway Neon project (created and torn down for the test) in a platform-served region, minting a real `ai_gateway:invoke` credential. Here's what each check proved, stated honestly:

**✅ Credential + gateway auth (verified live).** The documented credential endpoint returned `201` with an `api_token`; an unauthenticated `GET /v1/models` returned `401`, and the same call with `Authorization: Bearer <credential>` returned `200`. So the documented credential flow and bearer-auth requirement work exactly as written.

**✅ Rate-limit error text — the corrected string is right (verified live).** We drove a model past its per-minute token budget and the gateway returned `429` with the body message `ai gateway per-minute token limit exceeded for model "databricks-gemini-3-1-flash-lite"` — matching the corrected wording verbatim, and confirming the old "TPM limit exceeded" string is not what's emitted. **Narrow point:** we only ran an over-threshold probe (~210k tokens), so we proved a per-minute limit *exists and returns this exact message*, but did not pin the doc's specific `200,000` tokens/minute number (a calibrated just-under/just-over test would be needed for that). The wording fix — which is what this PR changes — is fully verified.

**⚠️ Generated env-var count — not yet verified (test hit the wrong surface).** The doc claim is about the *Console credential snippet / `.env` download*. Our harness tested `neon env pull` instead, which is a different surface, and it returned zero gateway vars. The independent grader flagged this as a surface mismatch: `env pull` returning no gateway vars may simply be its correct behavior, so this does **not** prove the doc wrong — it's *blocked pending a proper test*, not a fail. Verifying it needs the actual Console snippet/`.env` download to be inspected (or first confirming from source whether `env pull` is even supposed to emit those vars). The "all four → both" correction itself is still supported by the docs-vs-product audit and by the page's own env-var table.

All three checks were independently graded by a different AI tool (a different vendor than the one that ran the tests). The grader **agreed** the credential/auth and rate-limit-wording results are PASS, and **corrected** the env-var result from a proposed "fail" down to "blocked / needs the right test surface" — so nothing here is over-claimed.

## Notes
- Tracked in LKB-15122 (error text) and LKB-15124 (env var count). Cross-vendor reviewed: PASS. Live test: rate-limit wording PASS (per-minute message verified verbatim; exact 200k/min number not pinned), credential/auth PASS, generated-env-count blocked/unverified (retry needs the Console snippet/`.env` download surface).

> **Note on the diff delta:** the original audit finding was the doc claiming "**all four** gateway env vars." By the time this branch was rebased onto `main`, a sibling ai-gateway PR had already softened that line to "the gateway env vars," so the *mechanical* diff here reads "the" → "**both**" (one line). The final wording correctly reflects the two gateway env vars either way; only the size of the delta is smaller than the "all four → both" framing above implies.


philip (approved):
LGTM

PR #5210, opened 2026-07-07
docs(functions): bring docs up to parity with last 30 days of code + IA audit

bgrenon:
## Summary
Brings Functions docs up to parity with the last 30 days of `neon-cloud`/`neon-pkgs` changes, corrects 2 confirmed cases of doc-vs-code drift, and fixes findings from a persona-driven IA read-through.

## Changes
- **`overview.md`**: added an explicit JS/TS-only runtime callout
- **`deploy.md`**: documented config-only/partial deploys, the 2-minute build timeout, the multipart deploy request shape, and the `@neon/sdk` `functions.*` wrapper; fixed a deploy example using non-existent `--path`/`--entry` flags → now uses the real `--src` flag
- **`environment-variables.md`**: corrected two drift cases found after the fact —
  - `NEON_STORAGE_*` vars were renamed to AWS-standard `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_ENDPOINT_URL_S3`/`AWS_REGION` (`neon-cloud` #7879); `NEON_STORAGE_FORCE_PATH_STYLE` was dropped entirely, no replacement
  - `OPENAI_BASE_URL` now points at the unified `/v1` base for both Chat Completions and Responses (`neon-cloud` #8146); removed the stale "two separate AI Gateway endpoints" guidance
  - Swapped a user-secret example off the reserved `OPENAI_API_KEY` name (collides with the platform-injected var) to `MY_SECRET_KEY`
- **`authentication.md`**: added the missing `npm install jose` step ahead of the JWT verification example (first deploy would otherwise fail to build)
- **`public/docs/ai/skills/neon-functions/SKILL.md`**: added the 3 doc pages missing from "Further reading" (`agents.md`, `websockets.md`, `authentication.md`) — these contain most of the real code patterns

PR #5324, opened 2026-07-20
docs(functions): fix reserved-prefix claim and invocation URL trailing slash

bgrenon:
Doc-only fixes from the neon-docs-drift audit (docs vs. current product behavior). Two independent corrections across 3 files. **Not auto-merged** — awaiting your review.

### 1. False "reserved `NEON_` prefix" claim (`runtime-limits.md`)
No env-var name is actually reserved.

- **Before:** "…limits on variable count, total size, and the reserved `NEON_` prefix."
- **After:** "…limits on variable count and total size."
- **Why the new text is right:** platform variables (`DATABASE_URL`, `AWS_*`, `NEON_*`, …) are injected as *defaults* — a user variable with the same name simply overrides them, so nothing is reserved. You can confirm by setting a `NEON_`-prefixed variable on a function and seeing it accepted. This is the same fix already merged for the sibling pages in #5210; this is the one file that PR missed.

### 2. Invocation URL examples missing the trailing slash
The public API guarantees a trailing slash so paths concatenate cleanly.

- **Before:** `https://<branch_id>-<slug>.compute.<cell>.us-east-2.aws.neon.tech`
- **After:** `https://<branch_id>-<slug>.compute.<cell>.us-east-2.aws.neon.tech/`
- **Why the new text is right:** the `invocation_url` returned by the public API always ends with a trailing slash so request paths concatenate onto it — you can confirm by deploying a function and inspecting the `invocation_url` the API returns. Applied to all 3 examples (the URL line and the curl line in get-started, and the deploy example).

## Files
| File | Change |
|------|--------|
| `content/docs/compute/functions/reference/runtime-limits.md` | drop the false "reserved `NEON_` prefix" clause |
| `content/docs/compute/functions/get-started.md` | trailing slash on URL + curl examples |
| `content/docs/compute/functions/deploy.md` | trailing slash on invocation URL example |

Both fixes were tested against a real Neon Functions deployment (production, Postgres 17.10, nodejs24 runtime). Anyone can reproduce:

**The `NEON_` prefix is not reserved.** Deploy a function with an env var named `NEON_LIVE_TEST_VALUE`. What happens: the deploy is accepted, the variable shows up in the function's environment, and calling the function returns the exact value you set. So a `NEON_`-prefixed name is accepted *and* readable at runtime — nothing reserves it.

**The invocation URL ends with a trailing slash.** Deploy a function and read the `invocation_url` the API returns. What happens: it ends in `/` — confirmed for two function slugs, including the `hello` slug from the get-started example.

A second, independent reviewer (a different AI tool) checked that these tests actually prove the doc's claims — both PASS. Text-only change; no build or deploy behavior is affected.

## Notes
- Tracked in LKB-15130 and LKB-15131. Cross-vendor reviewed: PASS. Live-tested against production Neon: PASS (both claims).

### DIFF
--- a/content/docs/compute/functions/deploy.md
+++ b/content/docs/compute/functions/deploy.md
@@ -190,7 +190,7 @@ GET /projects/{project_id}/branches/{branch_id}/functions/{slug}
 The response includes `invocation_url`, the public URL for your function:
 
 ```
-https://<branch_id>-<slug>.compute.<cell>.us-east-2.aws.neon.tech
+https://<branch_id>-<slug>.compute.<cell>.us-east-2.aws.neon.tech/
 ```
 
 ### List functions
--- a/content/docs/compute/functions/get-started.md
+++ b/content/docs/compute/functions/get-started.md
@@ -169,13 +169,13 @@ neon functions get hello
 The `invocation_url` field contains the public URL for your function:
 
 ```
-https://<branch_id>-<slug>.compute.<cell>.us-east-2.aws.neon.tech
+https://<branch_id>-<slug>.compute.<cell>.us-east-2.aws.neon.tech/
 ```
 
 Call it with curl:
 
 ```bash shouldWrap
-curl https://<branch_id>-hello.compute.<cell>.us-east-2.aws.neon.tech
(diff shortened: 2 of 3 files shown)


philip (approved):
:shipit:

PR #5325, opened 2026-07-20
docs(cli): correct functions get output title and vpc endpoint --region-id examples

bgrenon:
Doc-only CLI fixes from the neon-docs-drift audit (docs vs. current CLI source). Two independent corrections across 2 files. **Not auto-merged** — awaiting your review.

### 1. Wrong heading in a `functions get` example (`functions.md`)
For a single deployment (the case the example shows), the CLI prints a different heading.

- **Before:** `active deployment`
- **After:** `deployment (current, active)`
- **Why the new text is right:** `packages/cli/src/commands/functions.ts:437-459` — when the current and active deployment are the same id (the single-deploy example), the title is `deployment (current, active)`. The `active deployment` heading only appears when current ≠ active.

### 2. VPC endpoint examples missing a required flag (`vpc.md`)
`--region-id` is required on all `vpc endpoint` subcommands, so the copy-paste examples would fail without it.

- **Before:** `neon vpc endpoint list --org-id org-bold-bonus-12345678`
- **After:** `neon vpc endpoint list --org-id org-bold-bonus-12345678 --region-id aws-us-east-1`
- **Why the new text is right:** `packages/cli/src/commands/vpc_endpoints.ts:43-47` sets `--region-id` as `demandOption: true` on the parent `endpoint` command, so it's required for list, remove, and status alike. Applied to all three examples (the `assign` example already had it).

## Files
| File | Change |
|------|--------|
| `content/docs/cli/functions.md` | example heading → "deployment (current, active)" |
| `content/docs/cli/vpc.md` | add `--region-id` to list / remove / status examples (3 lines) |

Both corrections were exercised against a live throwaway Neon project (created and torn down for the test) — not just checked against source.

**✅ `neon functions get` heading is `deployment (current, active)` (verified live).** We deployed a single function to a fresh project, then ran `neon functions get`. With one deployment, the current and active deployment are the same id (confirmed: `current_deployment.id == active_deployment.id == 1`, both `completed`), and the CLI printed the heading exactly as corrected: `deployment (current, active)`. That's the doc's claim, read back verbatim from the tool the doc names.

**✅ `neon vpc endpoint` commands require `--region-id` (verified live).** Run `neon vpc endpoint list` with no `--region-id` — it errors with `Missing required argument: region-id`. Add `--region-id aws-us-east-1` and the command runs. Same behavior confirmed for `list`, `status`, and `remove`. So the corrected examples that include `--region-id` are right.

Both checks were independently graded by a different AI tool (a different vendor than the one that ran the tests): both confirmed **PASS**, with the functions-heading result judged *directly observed* (the exact heading string was read back from `neon functions get`), not merely inferred from a successful deploy.

## Notes
- Tickets: LKB-15144 (heading), LKB-15145 (`--region-id`). Cross-vendor reviewed: PASS. Live test: both corrections PASS (functions heading verified in a platform-served region; `--region-id` requirement verified).


philip (approved):
:shipit:

PR #5326, opened 2026-07-20
docs(neon-ts): fix stale quoted dataApi compiler error string

bgrenon:
Doc-only fix from the neon-docs-drift audit (docs vs. current product behavior). One correction in one file. **Not auto-merged** — awaiting your review.

## What changed and why

A quoted TypeScript compiler error in the docs used a stale product name. The real error text says "Neon Auth", not "Managed Better Auth".

- **Before:** "…requires Managed Better Auth, so add `auth: true`…"
- **After:** "…requires Neon Auth, so add `auth: true`…"
- **Why the new text is right:** this is the actual message the TypeScript compiler emits — you can confirm by omitting the auth config and reading the resulting compiler error, which now reads "Neon Auth". Three occurrences inside the one fenced `text` error block were updated.

## Scope / what's intentionally untouched
- **Only** the verbatim compiler-error quote block changed. The surrounding prose that says "Managed Better Auth" as a product name (e.g. the lead-in sentence) is **left intact on purpose** — that's a correct product-name usage, not part of the quoted error.

## Files
| File | Change |
|------|--------|
| `content/docs/reference/neon-ts.md` | 3× "Managed Better Auth" → "Neon Auth", inside the error block only |

This fixes the TypeScript error message shown for an invalid `@neon/config` config. Verified by actually compiling the config:

**Invalid config is rejected with the corrected message.** Write a `neon.config.ts` that sets `dataApi: true` but omits `auth: true`, then run the TypeScript compiler. What happens: compilation fails on the `dataApi` line with the exact corrected error text — the one referencing **Neon Auth** (not the old "Managed Better Auth" wording). The two documented fixes — adding `auth: true`, or using the external-IdP object form — both compile cleanly.

A second, independent reviewer (a different AI tool) confirmed the compiler emitted the *full* corrected message verbatim, not just a fragment — PASS.

## Notes
- Tracked in LKB-15155. Cross-vendor reviewed: PASS. Live-tested against production Neon: PASS.

### DIFF
--- a/content/docs/reference/neon-ts.md
+++ b/content/docs/reference/neon-ts.md
@@ -173,9 +173,9 @@ Run `neon deploy` to apply. When `neon checkout` creates a new branch, the closu
 `dataApi: true` uses Managed Better Auth as the JWT verifier (the default). When using this form, `auth: true` must also be set. Omitting it raises a TypeScript error at the `dataApi` field that includes the fix:
 
 ```text
-Type 'true' is not assignable to type '"`dataApi` with Managed Better Auth (the default
-`authProvider: 'neon'`) requires Managed Better Auth, so add `auth: true`. To enable the
-Data API WITHOUT Managed Better Auth, verify a third-party IdP instead: `dataApi: {
+Type 'true' is not assignable to type '"`dataApi` with Neon Auth (the default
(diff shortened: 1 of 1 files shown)


philip (approved):
:shipit:

PR #55, opened 2022-06-01
docs/quickstart: clean up and fix broken links

yeputons:
* Clean up numbered lists like in #40 
* Fix some broken links
* Remove weird symbols in random places

PR #5360, opened 2026-07-25
chore: sync hosted agent skills with upstream main

andrelandgraf:
## Summary

Runs `npm run update:skills` (`sync:skills` + `generate:skills`) from latest `main` to bring the hosted agent skills up to date with `neondatabase/agent-skills@main`, and regenerates the discovery indexes and AI catalog. Nothing under `public/docs/ai/skills/` was hand-edited — it's all script output, per the read-only mirror rule.

All nine configured skills were re-synced (eight from `neondatabase/agent-skills`, plus `neon-postgres-agent-platforms` from `neondatabase/neon-for-agent-platforms`), picking up [agent-skills#63](https://github.com/neondatabase/agent-skills/pull/63):

- Every skill reworded and reordered for agent readability, with blocking preconditions (beta regions, paid-plan requirements) moved to the top.
- **CLI is now the documented default over MCP**; `neon-postgres-branches` previously told agents to check MCP first, contradicting the parent `neon` skill.
- Two factual corrections, both verified against live Neon: `neon functions deploy` takes `--src` (the documented `--path` / `--entry` don't exist), and `neon env pull` does not emit AI Gateway vars from branch state — `neon.ts` must declare `preview.aiGateway`.
- Frontmatter descriptions were corrected upstream, so `catalog.json` and the four index files pick up better routing text — notably `neon-postgres` (previously advertised Neon Auth and the Python SDK, which it doesn't cover) and `neon-functions` (now lists "MCP server" as a trigger).

## One change beyond the script output

`scripts/check-skills-sync.mjs` was **failing** before this PR:

```
[FAIL] neon-postgres — extra local file not in upstream: references/neon-sdk.md
```

`sync-skills.js` only downloads files; it never prunes ones deleted upstream. `neon-postgres/references/neon-sdk.md` was removed from the source repo, so the mirror had drifted and the check enforces a 1:1 match. This PR deletes it, and the check now reports **all 9 skills match upstream**.

That file was the static `.md` fixture for two tests, so both now point at `neon-functions/references/sentry.md`, which is vendored and satisfies the same `body contains "neon"` spot-check:

- `scripts/test-markdown-urls.js` — actually fetches the URL and asserts 200 + non-empty body, so it would have broken against a deployed site.
- `src/middleware.test.js` (2 occurrences) — only uses the path as a mocked string, but pointing a "static file passes through" assertion at a non-existent file would be misleading.

`@neon/sdk` guidance itself isn't lost: it now lives in the `neon` skill's "Manage Neon Resources" section.

## Why the diff shows a large deletion

`public/docs/ai/skills/neon-postgres/SKILL.md` drops ~296 lines because the hosted copy predated upstream [#55](https://github.com/neondatabase/agent-skills/pull/55), which slimmed that skill by replacing inlined prose with `neon.com` doc links. The `neon` skill grew over the same span. This is staleness catching up, not content removal.

## Test plan

- [x] `npm run update:skills` from latest `main` (`358a69129`)
- [x] `node scripts/check-skills-sync.mjs` — **[OK] All 9 skill(s) match upstream** (was failing before)
- [x] All 15 synced files diffed against `agent-skills@main` — byte-identical, no hand-edits
- [x] `npx vitest run src/scripts/generate-skills-index.test.js src/middleware.test.js` — 85 tests pass
- [x] `npx eslint` clean on both changed JS files
- [x] All five generated JSON files parse
- [ ] `scripts/test-markdown-urls.js` hits live URLs, so the repointed static path is best confirmed against a preview deploy


philip (approved):
Not much to review here as this is a simple sync, so review belongs upstream.

PR #5361, opened 2026-07-26
docs: Add DIRECT_URL in .env example and update drizzle kit config

darreleng:
Updated database connection strings for pooled and unpooled connections in the Drizzle documentation.


danieltprice:
Thanks for this — the pooled/unpooled split is the right call and matches the guide's own advice to use a direct connection for Drizzle Kit migrations.

I made two adjustments before merging and live-tested the result:
- **Standardized the unpooled var to `DATABASE_URL_UNPOOLED`** (instead of `DIRECT_URL`), to match the name Neon's console/Vercel integration actually provisions and the convention used across the rest of the docs.
- **Fixed the `drizzle.config.ts` guard** to validate `DATABASE_URL_UNPOOLED` (the variable the config uses) rather than `DATABASE_URL`, and aligned the `.env.local` variant too.

Live test: scaffolded this exact setup against a real Neon project — `drizzle-kit push` applied the schema via `DATABASE_URL_UNPOOLED`, and INSERT/SELECT/DELETE worked through the pooled `DATABASE_URL`. Merging.

danieltprice (approved):
Normalized to DATABASE_URL_UNPOOLED + fixed config guards; live-tested drizzle-kit push (unpooled) + CRUD (pooled) end to end.

PR #5364, opened 2026-07-27
docs(cli): point binary downloads and the docs refresh at the neon release

andrelandgraf:
## Problem

The Neon CLI is being renamed from `neonctl` to `neon` ([neon-pkgs#331](https://github.com/neondatabase/neon-pkgs/pull/331), [releases#187](https://github.com/databricks/secure-public-registry-releases-eng/pull/187)). Two things in this repo are coupled to the old name and neither fails loudly.

**1. The documented binary downloads.** `content/docs/cli/install.md` tells people to `curl` five URLs of the form `releases/latest/download/neonctl-<platform>`. After the rename the release ships `neon-<platform>` assets under a `neon@<version>` tag.

**2. The CLI reference pipeline.** `scripts/docs-checks/neonctl/refresh.js` picks the highest-versioned `neonctl@*` release to regenerate `schema.json`. Once releases move to `neon@*`, it doesn't error — it keeps resolving `neonctl@2.37.1` forever, so the entire CLI reference silently freezes at that version.

## Solution

Point both at the new names.

- The five `curl` commands and the Windows run example now use `neon-<platform>`.
- `TAG_PREFIX` is `neon@`. `neon-new@*` and friends don't collide, because the `@` has to follow `neon` immediately — the existing `startsWith` check is enough.

## The old URLs keep working

The release workflow uploads every binary under **both** names, so `releases/latest/download/neonctl-linux-x64` does not 404 — that rolling URL is baked into user CI (21 downloads on the 2.37.1 release alone), and the rename must not break it. This PR changes which name we *document*, not which ones exist.

That also means this PR is a no-op for correctness and purely a naming cleanup. Its only hard requirement is ordering.

## Merge order

**Merge after `neon@2.38.0` is published** — the `neon-*` assets and the `neon@*` tag don't exist until then. Draft until the release lands.

## Verification

- Grepped the repo for every reference to the release assets and tag prefix; these two files are the only ones. `npx neonctl@latest` references in blog posts and changelogs are npm installs, which are unaffected — `neonctl` stays published as a compatibility package.
- Asset names confirmed against the workflow's `pkg` output: `neon-linux-arm64`, `neon-linux-x64`, `neon-macos-x64`, `neon-win-x64.exe`.
- Couldn't run `prettier` locally (repo deps not installed, missing `prettier-plugin-tailwindcss`); changes are in-line string substitutions and comments, so CI formatting checks should be authoritative here.

## Follow-up

The `neonctlVersion` schema field and the `scripts/docs-checks/neonctl/` directory name still say `neonctl`. Left alone deliberately — renaming them touches the schema contract and is unrelated to this break.

### DIFF
--- a/content/docs/cli/install.md
+++ b/content/docs/cli/install.md
@@ -47,7 +47,7 @@ bun install -g neon
 Download the binary. No installation required.
 
 ```bash shouldWrap
-curl -sL https://github.com/neondatabase/neon-pkgs/releases/latest/download/neonctl-macos-x64 -o neon
+curl -sL https://github.com/neondatabase/neon-pkgs/releases/latest/download/neon-macos-x64 -o neon
 ```
 
 Run the CLI from the download directory:
@@ -79,13 +79,13 @@ bun install -g neon
 Download the binary. No installation required.
 
 ```bash shouldWrap
-curl -sL -O https://github.com/neondatabase/neon-pkgs/releases/latest/download/neonctl-win-x64.exe
+curl -sL -O https://github.com/neondatabase/neon-pkgs/releases/latest/download/neon-win-x64.exe
 ```
 
 Run the CLI from the download directory:
 
 ```bash
-neonctl-win-x64.exe <command> [options]
+neon-win-x64.exe <command> [options]
 ```
 
 </TabItem>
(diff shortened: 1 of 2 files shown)


philip (approved):
🚢 🇮🇹

PR #5366, opened 2026-07-27
fix(components): add missing observability category to MCP config generator

andrelandgraf:
## Problem

The MCP server added an `observability` tool category ([mcp-server-neon#292](https://github.com/neondatabase/mcp-server-neon/pull/292)). The **Available tools** table in `content/docs/shared-content/mcp-tools.md` was updated for it — the **config generator** on [/docs/ai/neon-mcp-server#config-generator](https://neon.com/docs/ai/neon-mcp-server#config-generator) was not.

This was worse than a cosmetic omission. `buildQueryParams` only emits `?category=` once the selection is a strict subset of the known list, so:

- With everything checked, no `category` param is emitted and the URL is correct — the gap is invisible.
- The moment a user unchecks **any** category, the generator emits a `category` list built from the 7 it knew about, **silently dropping `observability`** — disabling five tools the user never chose to turn off.

Before this PR, unchecking "Projects" produced:

```
https://mcp.neon.tech/mcp?category=branches,schema,querying,neon_auth,data_api,docs
```

After:

```
https://mcp.neon.tech/mcp?category=branches,schema,querying,neon_auth,data_api,observability,docs
```

## Solution

1. Add `observability` to `SCOPE_CATEGORIES` in the config generator, in the server's order (between `data_api` and `docs`).
2. Add a drift check so the next category can't rot the same way.

The MCP server advertises its own category list, so this is verifiable rather than a convention:

```bash
curl -s https://mcp.neon.tech/.well-known/oauth-authorization-server | jq '."x-neon-scope-categories"'
[ "projects", "branches", "schema", "querying", "neon_auth", "data_api", "observability", "docs" ]
```

`scripts/check-mcp-categories-sync.mjs` parses the two places this repo restates that list — `SCOPE_CATEGORIES` and the "Available tools" table — and compares them. Order is enforced too, since both are rendered to users. It follows the existing `agent-discovery-verify` / `skills-readonly` pattern:

- **Offline (PR gate):** generator and docs table must agree. No network, deterministic. This alone catches this exact bug, since the docs table was updated and the generator wasn't.
- **`--live` (daily 09:00 UTC + manual):** both are additionally compared against the server's advertised list, and drift opens/updates a tracking issue. This is what catches a new category that nobody mirrored here at all.

Zero-dependency (Node builtins + `fetch`), so the workflow needs no `npm ci`.

## User-facing API

New npm scripts:

```bash
npm run check:mcp-categories        # offline: generator vs docs table
npm run check:mcp-categories:live   # + compare both against mcp.neon.tech
```

`--server <base>` overrides the MCP base URL for staging.

Failure output names the drift and how to fix it:

```
  [FAIL] config generator vs MCP server
           - missing from config generator: observability
           MCP server: [projects, branches, schema, querying, neon_auth, data_api, observability, docs]
           config generator: [projects, branches, schema, querying, neon_auth, data_api, docs]

To fix, make all listed sources agree with the MCP server, in the same order:
  - config generator: SCOPE_CATEGORIES in
    src/components/pages/doc/mcp-setup-configurator/mcp-setup-configurator.jsx
    (each entry also needs a user-facing label and description)
  - docs table: the "Available tools" table in content/docs/shared-content/mcp-tools.md
```

## Verification

**The check is red before the fix, green after.** Stashing the `SCOPE_CATEGORIES` change reproduces the failure above in both modes (exit 1); with the fix both exit 0.

- `npm run check:mcp-categories` and `:live` — pass.
- `npm run test:unit:run` — 618 passed (33 files), including 10 new tests covering both parsers and the diff formatter (missing / extra / reordered / exact match, plus the throw paths for a renamed `SCOPE_CATEGORIES` or a missing table).
- ESLint + Prettier clean.

**Browser (local dev, real `mcp.neon.tech`):**

- All 8 categories render, light and dark mode.
- Selecting **only** Observability produces `?category=observability`, and the live tools preview returns exactly the expected 5 tools — `Search`, `Fetch`, `Query Logs`, `List Log Fields`, `List Log Field Values` — matching `curl "https://mcp.neon.tech/api/list-tools?category=observability"`.
- Unchecking one category now keeps `observability` in the generated URL (the regression above).

## Screenshots

The **Tool categories** card now renders a uniform 4×2 grid, all eight entries on a single description line, verified in both light and dark mode:

```
Projects              Branches
Create and manage…    Create, reset, delete…

Schema                Querying
Tables, columns…      Run SQL and explain plans

Neon Auth             Data API
Users and sessions    RESTful data endpoints

Observability         Docs
Function and…         Search and fetch docs
```

I can attach the PNGs on request — flagging that I verified visually rather than pasting images, since I generated them locally.

## Notes and known gaps

- **Description wording** — `observability` is described as "Function and storage logs" to match the one-line noun-phrase style of the neighbouring cells; a longer string wraps and makes that grid row taller than the others. Happy to reword.
- **`?category=` is comma-joined, docs show it repeated.** `buildQueryParams` emits `?category=a,b`; the docs show `?category=a&category=b`. I verified the server accepts both and returns identical grants, so this is not a bug and I left it alone — flagging it only because the generated URL doesn't look like the documented form.
- **The daily live check depends on `mcp.neon.tech` being reachable** from the runner. If the well-known endpoint is flaky it will open a drift issue; the offline PR gate is unaffected.
- **This only covers category *names*, not labels or descriptions.** The server no longer ships human-readable labels (they were removed in mcp-server-neon#199), so the generator's label/description text stays hand-written and unverified. Adding a category still requires writing those by hand — the check just makes it impossible to forget the entry entirely.

### DIFF
--- a/.github/workflows/mcp-categories-sync.yml
+++ b/.github/workflows/mcp-categories-sync.yml
@@ -1,0 +1,75 @@
+name: MCP Categories Sync
+
+# The Neon MCP server owns the list of tool categories (`?category=` values).
+# Two places here restate it for readers and silently rot when the server gains
+# or drops one: SCOPE_CATEGORIES in the docs config generator, and the
+# "Available tools" table in content/docs/shared-content/mcp-tools.md. A missing
+# entry in the generator is the worst case — it emits a `?category=` URL that
+# quietly disables tools the user never chose to turn off.
+#
+#   - Pull requests touching either surface -> offline check (no network): the
+#     generator and the docs table must agree.
+#   - Daily schedule / manual dispatch -> --live: both are additionally compared
+#     against mcp.neon.tech's advertised x-neon-scope-categories, and drift
+#     opens/updates a tracking issue so a new category can't rot silently.
+
+on:
+  pull_request:
+    paths:
+      - 'src/components/pages/doc/mcp-setup-configurator/**'
+      - 'content/docs/shared-content/mcp-tools.md'
+      - 'scripts/check-mcp-categories-sync.mjs'
+      - '.github/workflows/mcp-categories-sync.yml'
+  schedule:
+    # Every day at 9:00 AM UTC (after the agent-discovery and skills checks)
+    - cron: '0 9 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  check:
+    name: 'Check MCP tool categories are in sync'
+    runs-on:
+      group: neondatabase-protected-runner-group
+      labels: linux-ubuntu-latest
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@4d991eb9b905ef189e4c376166672c3f2f230481 # v2.11.0
+        with:
+          egress-policy: audit
+
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+        with:
+          node-version: '20'
+
+      # Zero-dependency check (Node builtins + global fetch) — no install needed,
+      # so nothing to fetch from the npm registry.
+      - name: Check categories (offline)
+        if: github.event_name == 'pull_request'
+        run: node scripts/check-mcp-categories-sync.mjs | tee mcp-categories-report.txt
+
+      - name: Check categories (live)
+        if: github.event_name != 'pull_request'
+        run: node scripts/check-mcp-categories-sync.mjs --live | tee mcp-categories-report.txt
+
+      - name: Open or update drift issue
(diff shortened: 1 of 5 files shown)


andrelandgraf:
### CI + preview verification

Both workflow paths now confirmed on real runners, not just locally:

- **Offline (PR gate)** — [run 30280251703](https://github.com/neondatabase/website/actions/runs/30280251703), pass in 51s, no `npm ci` needed.
- **Live (`--live`)** — [run 30280524660](https://github.com/neondatabase/website/actions/runs/30280524660), dispatched manually against this branch. Reached `mcp.neon.tech/.well-known/oauth-authorization-server` from the runner and matched both local lists. Confirms the daily schedule will work once merged.

On the [Vercel preview](https://neon-next-git-fix-mcp-config-generator-obse-826b29-neondatabase.vercel.app/docs/ai/neon-mcp-server): all 8 categories render, and unchecking "Projects" produces

```
https://mcp.neon.tech/mcp?category=branches,schema,querying,neon_auth,data_api,observability,docs
```

with `observability` retained — the regression this PR fixes.

guillaumervls (approved):

PR #5451, opened 2026-08-05
fix(docs): prevent masked pg_dump failures

abdlrhmn0x7e:
## Summary

- Validate that `DATABASE_URL` is configured.
- Quote connection strings containing shell metacharacters.
- Propagate `pg_dump` failures through the gzip pipeline.

## Why

Unquoted URL parameters can change shell execution, while a successful `gzip` process can mask database connection and dump failures.

## Validation

- Markdown lint
- Missing-secret and pipeline-failure checks


danieltprice:
Thanks! Merging with a small tweak to the explanatory note.

The code fixes are all good: the `$POSTGRES/$pg_dump` → `pg_dump` correction (that second block referenced an undefined variable), quoting `"$DATABASE_URL"`/`"$GZIP_NAME"`, and the `DATABASE_URL` guard. I live-tested a real `pg_dump | gzip`: with a failing dump and no pipefail it exits 0 and writes a broken 20-byte file; with pipefail it correctly exits non-zero.

One nuance: GitHub Actions already runs `bash` steps with `-eo pipefail` by default, so in this workflow the failure wasn't actually masked. Setting `pipefail` explicitly is still fine (self-documenting + robust to shell overrides), so I kept it, and reworded the note to say so accurately rather than implying the step would otherwise report false success.

danieltprice (approved):
Valid fixes (pg_dump typo, quoting, guard); reworded the pipefail note to reflect GHA's default. Live-tested the masked-failure mechanism.

PR #5484, opened 2026-08-07
docs: document upgradeWebSocket as the WebSocket path for Neon Functions

philip:
## Overview

Rework the WebSockets and SSE guide to document `upgradeWebSocket` as the single supported path, following review feedback. Fetch handlers use the root helper from `@neon/functions` (0.7.0+); Hono apps use the `@neon/functions/hono` route helper (0.9.0+).

## What changed

- Remove the `ws`/`upgrade`-export path and its examples, so there's one way to accept a connection.
- Lead the Hono section with the `@neon/functions/hono` helper (`onOpen`/`onMessage`/`onClose`), with auth as normal route middleware.
- Warn about middleware that rebuilds the response: `cors()` on an upgrade route returns HTTP 500 `websocket_upgrade_response_lost`.
- Restore the `neon` CLI 2.45.0+ requirement for local WebSocket testing.
- Rewrite the `LISTEN/NOTIFY` example on `upgradeWebSocket` and a `Set` of sockets.
- `cli/dev.md`: drop the "both patterns" line.

## Testing

Verified end-to-end against the deployed runtime: the root and Hono helpers handshake to `101`, and the `cors()` case returns `500 websocket_upgrade_response_lost`.

### DIFF
--- a/content/docs/cli/dev.md
+++ b/content/docs/cli/dev.md
@@ -43,4 +43,4 @@ neon dev --source ./functions/hello.ts --port 3000
 
 ## WebSockets
 
-`neon dev` serves WebSocket upgrades the way the deployed runtime does, so you can test real-time functions locally. Both patterns work: `upgradeWebSocket` from `@neon/functions` and an `upgrade` export. Connect a client to the dev server's `ws://localhost:<port>` URL. See [WebSockets and SSE](/docs/compute/functions/websockets) for the function-side code.
+`neon dev` serves WebSocket upgrades locally, so you can test real-time functions before you deploy. Both patterns work: `upgradeWebSocket` from `@neon/functions` and an `upgrade` export. Connect a client to the dev server's `ws://localhost:<port>` URL. See [WebSockets and SSE](/docs/compute/functions/websockets) for the function-side code.
--- a/content/docs/cli/dev.md
+++ b/content/docs/cli/dev.md
@@ -40,3 +40,7 @@ Serve one function on an explicit port (fails if the port is taken):
 ```bash
 neon dev --source ./functions/hello.ts --port 3000
 ```
+
+## WebSockets
+
+`neon dev` serves WebSocket upgrades the way the deployed runtime does, so you can test real-time functions locally. Both patterns work: `upgradeWebSocket` from `@neon/functions` and an `upgrade` export. Connect a client to the dev server's `ws://localhost:<port>` URL. See [WebSockets and SSE](/docs/compute/functions/websockets) for the function-side code.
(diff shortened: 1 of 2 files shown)


danieltprice:
Heads-up for this rework: neon-pkgs #400 (merged Sep 1) shipped a dedicated Hono binding, `@neon/functions/hono`, with a first-party `upgradeWebSocket` route helper that takes `onOpen`/`onMessage`/`onClose`:

```ts
import { upgradeWebSocket } from '@neon/functions/hono';

app.get('/ws', upgradeWebSocket(() => ({
  onOpen:    (_e, ws) => ws.send('welcome'),
  onMessage: (e, ws)  => ws.send(`echo: ${e.data}`),
  onClose:   ()       => {},
})));
```

The current diff documents calling the low-level `@neon/functions` `upgradeWebSocket(c.req.raw)` inside a Hono route (good), but it drops the old pointer to the `onOpen`/`onMessage`/`onClose` helper. Worth adding `@neon/functions/hono` as the ergonomic Hono path — it replaces the deprecated `@hono/node-server` route helper. Happy to add it if useful.

andrelandgraf (changes requested):
Andre's agent: Philip, this updates the review. Ignore the earlier ask to keep the `{ fetch, upgrade }` + `ws` path.

Requesting changes. Document one WebSocket path only: `upgradeWebSocket`, with its Hono helper for Hono apps.

Delete the `{ fetch, upgrade }` + `ws` sections from `content/docs/compute/functions/websockets.md` and the "both patterns" sentence from `content/docs/cli/dev.md`. Remove all remaining mentions of the old path in this PR, including alternatives and "lower level" guidance.

For fetch handlers, document the root helper with a requirement of `@neon/functions` **0.7.0+**:

```ts
import { upgradeWebSocket } from '@neon/functions';

const { socket, response } = upgradeWebSocket(request);
return response;
```

Lead the Hono section with the Hono helper, as Daniel Price requested on 2026-09-04. Require `@neon/functions` **0.9.0+**:

```ts
import { upgradeWebSocket } from '@neon/functions/hono';

app.get('/ws', upgradeWebSocket(() => ({
  onOpen: (_event, ws) => ws.send('welcome'),
  onMessage: (event, ws) => ws.send(`echo: ${event.data}`),
  onClose: () => console.log('client disconnected'),
})));
```

Remove `upgradeWebSocket(c.req.raw)` as the documented Hono path.

Two requirements from the first review remain:

- Name the Hono middleware patterns that break the upgrade: `cors()`, reading `c.res` before `next()` and setting headers afterward. The reproduced case is `app.use('*', cors())` on an upgrade route returning HTTP 500 with `websocket_upgrade_response_lost`.
- Restore the live-docs requirement of `neon` **2.45.0 or later** for local WebSocket support. Installing `@neon/functions` does not upgrade the CLI.

PR #5489, opened 2026-08-08
fix: prune deleted files when syncing skills, and re-sync with upstream

### LINKED ISSUE #5460: Skills drifted from upstream agent-skills
Skills read-only / upstream drift check
  skills checked: 9

  [OK]   neon-postgres  (1 file(s), neondatabase/agent-skills@main:skills/neon-postgres)
  [OK]   claimable-postgres  (1 file(s), neondatabase/agent-skills@main:skills/claimable-postgres)
  [OK]   neon-postgres-egress-optimizer  (1 file(s), neondatabase/agent-skills@main:skills/neon-postgres-egress-optimizer)
  [OK]   neon-postgres-branches  (1 file(s), neondatabase/agent-skills@main:skills/neon-postgres-branches)
  [OK]   neon  (1 file(s), neondatabase/agent-skills@main:skills/neon)
  [OK]   neon-postgres-agent-platforms  (28 file(s), neondatabase/neon-for-agent-platforms@main:skills/neon-postgres-agent-platforms)
  [OK]   neon-object-storage  (1 file(s), neondatabase/agent-skills@main:skills/neon-object-storage)
  [FAIL] neon-ai-gateway  (neondatabase/agent-skills@main:skills/neon-ai-gateway)
           - content differs from upstream: SKILL.md
  [FAIL] neon-functions  (neondatabase/agent-skills@main:skills/neon-functions)
           - content differs from upstream: SKILL.md
           - content differs from upstream: references/sse.md
           - extra local file not in upstream: references/hono-websockets.md

— Run: https://github.com/neondatabase/website/actions/runs/31089193082

andrelandgraf:
Two things, in dependency order: fix `sync-skills.js` so it deletes vendored files that upstream removed, then re-sync. The fix is what makes the sync complete without a hand edit.

Closes #5460.

## Problem

The nightly `Skills Read-Only` check has failed on `main` since Aug 6 and filed #5460:

```
[FAIL] neon-ai-gateway  - content differs from upstream: SKILL.md
[FAIL] neon-functions   - content differs from upstream: SKILL.md
                        - content differs from upstream: references/sse.md
                        - extra local file not in upstream: references/hono-websockets.md
```

The first three lines are ordinary staleness. The last one could not be fixed by the remedy the check itself prints. `npm run update:skills` downloads every upstream file and writes it, and does nothing else — so `references/hono-websockets.md`, which upstream deleted in neondatabase/agent-skills#78 when its content was folded into `references/sse.md`, survived every re-sync and kept failing the "extra local file not in upstream" rule. The only way to clear it was a manual `rm`, which is exactly the hand edit the check exists to forbid.

That's a permanent trap, not a one-off: every future upstream deletion reproduces it.

## The fix

`syncSkill` now prunes after downloading. It already has the upstream file listing, so the local tree is compared against it and anything with no upstream counterpart is deleted, along with any directory left empty.

The guard that matters is the one that was already there. An empty upstream listing is refused **before** pruning, so an upstream rename, a moved `skills/` path, or a partial API response cannot delete a vendored copy:

```js
if (files.length === 0) {
  console.warn(`  Warning: no files found for skill "${name}" at ref "${ref}"`);
  return;
}
```

Output names every deletion and its reason, so a sync is auditable:

```
Syncing neon-functions @ main from neondatabase/agent-skills...
  ✓ SKILL.md
  ✓ references/sse.md
  - references/hono-websockets.md (removed upstream)
  - references/nested/deeper/orphan.md (removed upstream)
  - references/nested/deeper/ (empty)
  - references/nested/ (empty)
  Done. 6 file(s) synced, 2 deleted.
```

The deletion decision is a pure function, exported and unit tested:

```js
function filesToPrune(localRelativePaths, upstreamRelativePaths)  // → stale paths, sorted
function toLocalRelative(upstreamPath, skillName, skillsPath)     // → skill-relative path
```

`main()` is now gated behind `require.main === module`, matching `generate-skills-index.js`, so importing the module for those helpers doesn't start a sync.

## The re-sync

**`neon-postgres`** — the reason this happens now. neondatabase/agent-skills#83 added a `## Migrations` section and a `## Gotchas` entry telling agents to use the direct (non-pooled) connection string for schema migrations, `pg_dump`/`pg_restore`, and logical replication, and to test a migration on a branch of production first. It's the skill-side counterpart to the docs guidance in #5482 and uses @danieltprice's wording from that work. Frontmatter gains `schema migrations` in Covers and as a trigger.

**`neon-ai-gateway`, `neon-functions`** — undelivered upstream changes, unrelated to the above, including the `hono-websockets.md` deletion that the fix now handles.

**Five generated index files** — `public/.well-known/{agent-skills,skills}/index.json`, their `public/docs/` counterparts, and `src/app/.well-known/ai-catalog.json/catalog.json`, regenerated by `npm run generate:skills`. They embed frontmatter descriptions, so the `neon-postgres` description change propagates.

## Verification

Behaviour, not just green tests. `hono-websockets.md` was restored from `main` and two orphans planted at different depths, then `sync:skills` was run — it deleted all three and collapsed both emptied directories, output above. A second full `update:skills` run afterwards deleted nothing and left the vendored files untouched, so the operation is idempotent.

```
$ npx vitest run src/scripts/
 Test Files  5 passed (5)
      Tests  97 passed (97)

$ npm run check:skills-sync
  skills checked: 9
[OK] All 9 skill(s) match upstream.
```

The 8 new cases cover: local matching upstream, a stale file, a file upstream has that local doesn't (never pruned), several stale files returned sorted, an empty upstream listing, and the path transform including a per-skill `path` override.

Not verified: pruning has only been exercised against `neondatabase/agent-skills`, not against the `neon-postgres-agent-platforms` override that sources from `neon-for-agent-platforms`. Its 28 files synced clean in the full run, and `toLocalRelative` is covered for a custom `skillsPath`, but no deletion has actually run through that path.

## For your attention

- **Pruning is destructive by design.** The empty-listing guard is the only thing standing between an upstream oddity and a deleted vendored copy. It's load-bearing; keep it in place if this function is refactored.
- The sync pulls from upstream `main` rather than a pinned ref, so the re-synced content is a snapshot of `agent-skills@main` at the time it ran.


mezotv (approved):
LGTM!

PR #5570, opened 2026-08-18
perf(docs): defer mobile nav tree render to speed up static generation

rishi-raj-jain:
This made the static generation speed go from ~59s to ~37s across 2,328 pages!

The docs mobile-nav drawer is a JS-only overlay, hidden until tapped, whose links are already server-rendered in the desktop sidebar. It rendered the full ~1,700-node navigation tree into every prerendered page which was like ~35% of static generation time.

I deferred building/rendering that tree until the drawer is first opened. No visible and SEO change as desktop sidebar still carries the links.

### DIFF
--- a/src/components/pages/doc/mobile-nav/mobile-nav.jsx
+++ b/src/components/pages/doc/mobile-nav/mobile-nav.jsx
@@ -215,20 +215,29 @@ RecursiveList.propTypes = {
 
 const MobileMenu = ({ navigation, basePath, title = 'Neon Docs' }) => {
   const [open, setOpen] = useState(false);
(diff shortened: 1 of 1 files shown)


rishi-raj-jain:
Do we know how much time did the build step (in the deployment logs) of this vercel deployment took? and compare it across the production deployment build step (in the deployment logs).

rishi-raj-jain:
Update:

Me & @philip saw that this PR had a ~14s time on building the 2328 static pages while the production deployment has ~21s and other preview deployments have ~25s. Clearly, this approach helps us cut down the build times.

philip (approved):

PR #5612, opened 2026-08-24
docs: add Intellrise guide

MiyaN-ishere:
Hi, I'm from Intellrise, a conversational analytics tool for Lakebase Postgres. This adds a guide for connecting a Neon project to it, in the **Query** group alongside the existing entries.

### What is in the PR

- `content/docs/guides/intellrise.md`, a new page following the shape of the neighboring `askyourdatabase` and `draxlr` guides.
- `content/docs/navigation.yaml`, one entry in the Query group, alphabetically between Hasura and PostgREST.
- `content/docs/guides/integrations.md`, one card in the `## Query` section. It reuses `icon="database"`, the value the AskYourDatabase card already uses, so this PR ships no icon asset and touches nothing in `src/`.

Two notes on conventions, in case they save you a review round: I seeded `updatedOn` because every other page in the directory has it, but CLAUDE.md says not to set it by hand, so please let the pre-commit hook restamp it. And I could not run `npm run check:md` locally, so I hand-matched `.prettierrc`: LF endings, one paragraph per line with no reflowed prose, table cells padded to the column width, 3-space continuation under ordered items. If the format check still complains, tell me and I will fix it rather than leave it to you.

### No screenshots in this revision, and I would rather say so than pad the page

The neighboring guides carry product UI screenshots and I think that is the right format for this directory. This revision is text only: our connect screen changed this week, and shipping screenshots I cannot vouch for seemed worse than shipping none. Happy to push them to this branch before you merge if you would like them.

### On accuracy

Please treat this as a vendor-authored page and hold it to whatever bar that implies. Every product behavior claim was checked against our own source rather than written from our marketing copy, and that pass changed the draft in four places rather than confirming it. The parts that are not flattering are in the page on purpose:

- **There is no box to paste a Neon connection string into.** The form asks for host, port, database, schema, username and password separately. Rather than skip past that, the page has a table mapping each part of a Neon string onto the field it belongs in, since that is where a Neon user actually gets stuck.
- **Our read-only guarantee is a keyword filter, not a SQL parser**, and we do not check whether the account you enter is read-only. The page says both and recommends a dedicated read-only role, plus pointing at a branch instead of primary.
- **We connect from our servers, not the browser.** I gave this its own section because it is the thing most likely to waste a Neon reader's afternoon: with IP Allow enabled, the connection fails with a timeout that reads like a wrong hostname. I deliberately did not hardcode our egress ranges here, since they would go stale in your repo. The page points at the screen that prints the current ones.
- **A failed connection needs the password re-entered** when you correct it, which is a rough edge in our UI rather than something to be proud of. The page says so plainly instead of describing a flow that silently does nothing.

I also cut a paragraph about our handling of `channel_binding=require` that was in my first draft. It is real, but it only fires for a connection string pasted in whole, and this form never takes one, so documenting it would have implied an affordance the reader does not have.

If you would rather this page live under `content/guides/`, or not at all, that is a fine answer and I will not resubmit.

### DIFF
--- a/content/docs/guides/integrations.md
+++ b/content/docs/guides/integrations.md
@@ -117,6 +117,8 @@ updatedOn: '2026-08-04T08:34:18.168Z'
 
 <a href="/docs/guides/hasura" title="Hasura" description="Connect from Hasura Cloud to Neon" icon="hasura"></a>
 
+<a href="/docs/guides/intellrise" title="Intellrise" description="Query Lakebase Postgres in plain English with Intellrise" icon="database"></a>
+
 <a href="/docs/guides/postgrest" title="PostgREST" description="Create a REST API from your Neon database" icon="postgresql"></a>
 
 <a href="/docs/guides/wundergraph" title="Wundergraph" description="Use Wundergraph with Neon" icon="wundergraph"></a>
--- a/content/docs/guides/intellrise.md
+++ b/content/docs/guides/intellrise.md
@@ -1,0 +1,81 @@
+---
+title: Query Lakebase Postgres in plain English with Intellrise
+subtitle: Connect a Neon database and ask questions in English instead of SQL
+summary: >-
+  Intellrise is a conversational analytics tool that connects to a Lakebase
+  Postgres database, writes the SQL for a question asked in plain English,
+  runs it, and returns a chart or a table. Use this guide to connect a Neon
+  project to Intellrise, including how to map each part of a Neon connection
+  string onto the fields Intellrise asks for separately. The guide also
+  covers what Intellrise reads from your database and what it does not, the
+  network access it needs from a Neon project, and the read-only Postgres
+  role worth creating before you point it at production.
+enableTableOfContents: true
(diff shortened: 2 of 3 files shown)


MiyaN-ishere:
A week on, so a short note rather than a nudge — two things that might make this cheaper to decide either way.

**Screenshots are no longer a blocker.** The PR shipped without them and said why. I can add the two the neighbouring guides carry — the connection form and a first chart — whenever you want them. Say the word and I will push them, rather than leaving you to ask for something I already know is missing.

**If `content/docs/guides/` is the wrong door, I would rather move it than leave it sitting in your queue.** `content/guides/` takes a lighter review and has a `canonical` frontmatter field, which CLAUDE.md describes as being for exactly this case — a partner who also published the piece on their own site. That is where this would go if you prefer it, and honestly it is the better outcome for both of us.

A one-line "move it" or "not for us" is a perfectly good answer. Either way this is my only follow-up — I will not keep bumping the thread.

PR #5629, opened 2026-08-24
docs: sync with upstream skills

philip:


philip:
Closes https://github.com/neondatabase/website/issues/5614

PR #5658, opened 2026-08-27
docs: update @neon/sdk examples for 3.0 create / createAndConnect

andrelandgraf:
## Problem

Copy-paste samples on the TypeScript SDK reference, the api-client migration guide, the SDK blog post and the Cloudflare sandbox guide call `neon.branches.createWithCompute`. `@neon/sdk` 3.0.0 doesn't have that method. A reader on 3.0.0 fails at the call.

The Eve guide still follows `branches.create` with `postgres.endpoints.create({ type: "read_write" })`. On 3.0.0 `branches.create` already attaches a read-write endpoint.

The site's own dependency was `@neon/sdk@^2.0.0`.

## Diagnosis

`@neon/sdk` 3.0.0 splits the job by return value. Website copy was written for 2.x. Compute-plus-connection-string lived on `createWithCompute`. `branches.create` was branch-only.

3.0.0:

- `create` returns the resource. `projects.create` always attaches default-branch compute. `branches.create` attaches a read-write endpoint unless you pass `noCompute: true`. Both turn readiness polling on for that call.
- `createAndConnect` polls, then returns `{ project, connectionString }` or `{ branch, endpoint, connectionString }`.
- Input field names differ. `create` uses REST names (`parent_id`). `createAndConnect` uses `{ name?, parentId?, compute? }`.
- The client-wide `waitForReadiness` default stays `false`. Other mutations still pass `{ waitForReadiness: true }`.

A follow-up `endpoints.create` after `branches.create` is a second read-write endpoint. The 2.x `createWithCompute` workflow is `createAndConnect`.

## The user-facing interface

This PR updates neon.com samples and the site's `@neon/sdk` pin. The npm package already shipped.

```ts
const { data: branch, error } = await neon.branches.create(projectId, {
  name: "preview",
});
if (error) throw error;

const { data, error: connectError } = await neon.branches.createAndConnect(
  projectId,
  { name: "preview-uri" }
);
if (connectError) throw connectError;
const { connectionString } = data;
```

Branch reference, three modes (`create` uses `parent_id`; `createAndConnect` uses `parentId`):

```ts
await neon.branches.create(projectId, {
  name: "preview/pr-123",
  parent_id: prod?.id,
});

await neon.branches.create(projectId, {
  name: "schema-only",
  parent_id: prod?.id,
  noCompute: true,
});

const { data, error } = await neon.branches.createAndConnect(projectId, {
  name: "preview/pr-123-uri",
  parentId: prod?.id,
  compute: { minCu: 0.25, maxCu: 2 },
});
if (error) throw error;
const { connectionString } = data;
```

Project helper:

```ts
const { data, error } = await neon.projects.createAndConnect({
  name: "my-app",
});
if (error) throw error;
const { project, connectionString } = data;
```

Migration guide, branch with compute (returns `Branch`):

```ts
const { data, error } = await neon.branches.create(projectId, {
  name: "dev-1",
  parent_id: parentBranchId,
});
if (error) throw error;
const branch = data;
```

`createProjectBranch` maps to `branches.create` or `branches.createAndConnect`.

Cloudflare sandbox guide (`throwOnError: true`):

```ts
const { connectionString } = await neon.branches.createAndConnect(projectId, {
  name: branchName
})
```

Eve guide keeps `branches.create` plus `postgres.connectionString` and removes the extra `endpoints.create`.

| Method | Returns |
| --- | --- |
| `projects.create` | `Project` |
| `projects.createAndConnect` | `{ project: Project, connectionString: string }` |
| `branches.create` | `Branch` (read-write compute unless `noCompute: true`) |
| `branches.createAndConnect` | `{ branch, endpoint, connectionString }` |

Pagination, retries, `raw` and the other namespaces are unchanged.

## Also in here

- `package.json`, `package-lock.json` and `scripts/deps-proxy-allowlist.json` pin `@neon/sdk` at `3.0.0`.
- The branches method table drops `recover`. `@neon/sdk` 2.0.0 already removed `neon.branches.recover` from the ergonomic client. `projects.recover` stays.
- `updatedOn` on the four content files.
- The lockfile drops a `dev` flag on optional `fsevents`; that came with the install.

## Verification

- The files this PR edits no longer call `createWithCompute`.
- `package.json` is `@neon/sdk@^3.0.0`. `scripts/deps-proxy-allowlist.json` is `@neon/sdk@3.0.0`.
- Reference tables and samples use the 3.0.0 return types: `create` returns the resource; `createAndConnect` returns `{ project, connectionString }` or `{ branch, endpoint, connectionString }`.

Not verified: compiling the snippets against `node_modules/@neon/sdk@3.0.0` in this worktree (no install here). Playwright doesn't typecheck Markdown samples. Pages not rendered in the browser.

`rg createWithCompute` still has hits. See below.

## For your attention

- `public/docs/ai/skills/neon-postgres-agent-platforms/SKILL.md` still names `branches.createWithCompute` as a `@neon/tools` selector, in prose and in two `tools:` arrays. `check:skills-sync` requires that file to match upstream 1:1; a hand-edit in this repo fails CI. `check:docs-api-consistency` matches `neon.branches.createWithCompute(` (an SDK client call). The skill uses `"branches.createWithCompute"` tool ids, so that check stays green. The fix is an upstream skill refresh, then re-sync.
- The migration sample titled "Create a branch with compute" now returns `Branch`. Callers who copied it for `{ branch, endpoint, connectionString }` need `createAndConnect`; the mapping table lists both.
- `create` vs `createAndConnect` disagree on `parent_id` / `parentId`. Copying fields across the two calls will not typecheck.
- `branches.recover` is dropped from the reference table. The 2.0.0 changelog documents a raw `client.post` path for that endpoint.

### DIFF
--- a/content/blog/posts/neon-sdk.md
+++ b/content/blog/posts/neon-sdk.md
@@ -49,7 +49,7 @@ Most importantly, you can install `@neon/sdk` today:
 npm install @neon/sdk
 ```
 
-And use ergonomic functions like `createAndConnect` to provision a Neon project, wait for the provisioning operations to finish and hand you back a ready-to-use Postgres connection string, all in a single call:
+And use ergonomic functions like `createAndConnect` to provision a Neon project, wait for the provisioning operations to finish, and hand you back `{ project, connectionString }` in a single call:
 
 ```ts
 import { createNeonClient } from "@neon/sdk";
@@ -79,7 +79,7 @@ import { createNeonClient, raw } from "@neon/sdk";
 - **`createNeonClient`**: the high-level ergonomic client, organized into resource namespaces like `neon.projects` and `neon.branches`.
 - **`raw`**: the low-level generated surface, every endpoint as a standalone, tree-shakeable function.
 
-Take readiness polling as an example. The Neon API provisions real infrastructure and a lot of that work happens in the background, so most mutations don't hand you a ready-to-use resource. They return `operations` that you'd normally poll yourself until the resource is ready. `@neon/sdk` offers an abstraction that does that polling behind the scenes, so a call only resolves once the resource is actually ready. `createAndConnect` does this by default and you can opt any other mutation into the same behavior.
+Take readiness polling as an example. The Neon API provisions real infrastructure and a lot of that work happens in the background, so most mutations don't hand you a ready-to-use resource. They return `operations` that you'd normally poll yourself until the resource is ready. `@neon/sdk` offers an abstraction that does that polling behind the scenes, so a call only resolves once the resource is actually ready. `projects.create`, `branches.create`, and both `createAndConnect` workflows poll by default. Other mutations take `{ waitForReadiness: true }`.
 
 We hope `createNeonClient` covers most of what you need to build platforms and programmatic automations on Neon: a type-safe, ergonomic way to use Neon's capabilities. That said, you can always reach for the raw API methods directly.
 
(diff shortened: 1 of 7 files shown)


philip (approved):
:shipit:

PR #5662, opened 2026-08-27
Show OpenAI SDK examples for gpt-oss and qwen35

andrelandgraf:
## Problem

The latest chat-completions measurement records `gpt-oss-20b`, `gpt-oss-120b`, and `qwen35-122b-a10b` as conforming: `message.content` is a string and `deviations` is empty.

The site still classified all three as `array-content`. `/models` and neon.com/models therefore omitted their TypeScript and Python OpenAI SDK examples. The docs warning also listed two of these models as returning array content.

## Diagnosis

The example selector excludes OpenAI SDK snippets when `capabilities.chat` is `array-content`, because those SDKs type `message.content` as a string. The stale capability records activated that branch for these three models.

The refreshed measurement still records array content for Gemini 3.x and Claude 5. Those are now the model groups named in the warning.

## Solution

- Mark `gpt-oss-20b`, `gpt-oss-120b`, and `qwen35-122b-a10b` as `conforms` with no deviations.
- Verify that `/models` and the model page expose AI SDK, Mastra, TypeScript, Python, and cURL examples for each model.
- Verify that every remaining array-content model has a provider-native fallback.
- Retarget the docs warning at Gemini 3.x and Claude 5.
- Refresh the recorded Gemini token-accounting deviations from the same measurement.

The `/models` response structure and existing example templates remain unchanged. These models now select the conforming example set.

## User-facing API/UI

`GET /models?model=gpt-oss-20b` now includes these relevant fields:

```json
{
  "use_case": "chat",
  "model": {
    "id": "gpt-oss-20b",
    "capabilities": {
      "entitled": true,
      "chat": "conforms",
      "deviations": [],
      "native_dialect": "none",
      "responses": false,
      "web_search": false,
      "image_generation": false
    },
    "examples": [
      { "id": "ai-sdk", "endpoint": "/v1/chat/completions (via provider)" },
      { "id": "mastra", "endpoint": "/v1/chat/completions" },
      { "id": "typescript", "endpoint": "/v1/chat/completions" },
      { "id": "python", "endpoint": "/v1/chat/completions" },
      { "id": "curl", "endpoint": "/v1/chat/completions" }
    ]
  }
}
```

The same example set applies to `gpt-oss-120b` and `qwen35-122b-a10b`. Their model pages now offer AI SDK, Mastra, TypeScript, Python, and cURL in the language dropdown.

The docs warning ships as:

````mdx
<Admonition type="warning" title="Content shape varies by model">
For most models, `message.content` in a chat completions response is a plain string. For some models, confirmed on Gemini 3.x (`gemini-3-5-flash`, `gemini-3-1-pro`) and Claude 5 (`claude-sonnet-5`, `claude-opus-5`, `claude-fable-5`), it's an array of typed content blocks instead (`{ type: 'reasoning', ... }`, `{ type: 'text', text: ... }`). A low `max_tokens` value can also cut a response off before the `text` block appears, leaving only a `reasoning` block. Handle both shapes:

```typescript
const { content } = response.choices[0].message;
const text = typeof content === 'string'
  ? content
  : content.find((block) => block.type === 'text')?.text ?? '';
```

</Admonition>
````

## Usage examples

```typescript
import OpenAI from "openai";

const client = new OpenAI({
  apiKey: process.env.NEON_AI_GATEWAY_TOKEN,
  baseURL: `${process.env.NEON_AI_GATEWAY_BASE_URL}/v1`,
});

const resp = await client.chat.completions.create({
  model: "gpt-oss-20b",
  messages: [{ role: "user", content: "Explain Serverless Postgres." }],
});
console.log(resp.choices[0].message.content);
```

```python
import os
from openai import OpenAI

client = OpenAI(
    api_key=os.environ["NEON_AI_GATEWAY_TOKEN"],
    base_url=f"{os.environ['NEON_AI_GATEWAY_BASE_URL']}/v1",
)

resp = client.chat.completions.create(
    model="gpt-oss-20b",
    messages=[{"role": "user", "content": "Explain Serverless Postgres."}],
)
print(resp.choices[0].message.content)
```

Replace the model ID with `gpt-oss-120b` or `qwen35-122b-a10b` for the other two models.

## Verification

- Vitest against `route.test.js`, `model-examples.test.js`, `generate-ai-gateway-model-markdown.test.js`, and `model-rows.test.js`: 42 passed.
- `npm run check:models-endpoint-sync`: 42 models in sync.
- Pre-push hook: full unit suite passed (787 tests).

## Gaps

Live gateway output and the deployed model pages were not rechecked from this website worktree. The update relies on the committed conformance measurement and automated route/UI coverage. Gemini 3.x and Claude 5 still return array `content` on `/v1/chat/completions`; their OpenAI SDK examples stay hidden from the language dropdown.

### DIFF
--- a/content/docs/ai-gateway/models.md
+++ b/content/docs/ai-gateway/models.md
@@ -7,7 +7,7 @@ summary: >-
   Use short model IDs like gpt-5-mini or gemini-3-flash. The databricks- prefix
   is also accepted.
 enableTableOfContents: true
-updatedOn: '2026-08-17T22:56:05.062Z'
+updatedOn: '2026-08-27T22:18:16.922Z'
 ---
 
 <FeatureBetaProps feature_name="Neon AI Gateway" />
@@ -72,7 +72,7 @@ All paths below are appended to your branch's bare AI Gateway host (`NEON_AI_GAT
 | Meta, Alibaba, Zhipu AI, Thinking Machines, Moonshot AI | `/v1/chat/completions` | Chat completions only                                                                        |
 
 <Admonition type="warning" title="Content shape varies by model">
-For most models, `message.content` in a chat completions response is a plain string. For some models, confirmed on Gemini 3.x (`gemini-3-5-flash`, `gemini-3-1-pro`), `gpt-oss-120b`, and `qwen35-122b-a10b`, it's an array of typed content blocks instead (`{ type: 'reasoning', ... }`, `{ type: 'text', text: ... }`), matching how those models represent output natively. A low `max_tokens` value can also cut a response off before the `text` block appears, leaving only a `reasoning` block. Handle both shapes:
+For most models, `message.content` in a chat completions response is a plain string. For some models, confirmed on Gemini 3.x (`gemini-3-5-flash`, `gemini-3-1-pro`) and Claude 5 (`claude-sonnet-5`, `claude-opus-5`, `claude-fable-5`), it's an array of typed content blocks instead (`{ type: 'reasoning', ... }`, `{ type: 'text', text: ... }`). A low `max_tokens` value can also cut a response off before the `text` block appears, leaving only a `reasoning` block. Handle both shapes:
 
 ```typescript
 const { content } = response.choices[0].message;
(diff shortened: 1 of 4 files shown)


philip (approved):
Tested, all looks good 👍

PR #5676, opened 2026-08-30
docs: add Credentials & access concept page (LKB-16595)

bgrenon:
Adds a page, **Credentials & access**, to the new Concepts section. It explains how access to a Neon backend works as one model.

## Preview

https://neon-next-git-lkb-16595-credentials-access-neondatabase.vercel.app/docs/concepts/credentials-access

## Scope

It's a concept page, so it stays thin: reference tables, scope lists, and procedures stay in the Platform and per-product sections, and this page links down to them.

Also adds a **Backend behavior** section under Concepts in the sidebar for the page to live in.

## Concepts section (end state)

```
Concepts
├─ Backend primitives
│   ├─ Overview                 (planned)
│   ├─ The object model         (LKB-16593, base branch)
│   └─ Branch your backend      (planned · LKB-16594)
└─ Backend behavior
    ├─ Credentials & access     ← this PR (LKB-16595)
    ├─ Regions & availability   (planned · LKB-16600)
    └─ Serverless & scaling     (planned · LKB-16600)
```

## For the reviewer

Stacked on the Concepts-lane branch rather than `main`, so the diff shows only the new page plus the navigation entry. It retargets to `main` once that branch lands.

LKB-16595


content/docs/concepts/credentials-access.md
```diff
@@ -0,0 +1,138 @@
...
+  and its descendants, and it appears as an S3-compatible access key or a bearer
+  token. The Data API instead takes a JWT from a trusted issuer and runs each
+  query as a Postgres role. Your access to a project is your organization role
+  plus any per-project grant. IP Allow and Private Networking gate the Postgres
+  endpoint, and the other services are gated by credentials.
+enableTableOfContents: true
+---
+
+Neon issues two kinds of credentials, and they do different jobs:
+
```
danieltprice, line 17: We mention two types of credentials but the table that follows shows four.

content/docs/concepts/credentials-access.md
```diff
@@ -0,0 +1,138 @@
...
+- **Platform API keys** manage your Neon setup: create projects and branches, change settings, read usage, and issue the service credentials your app uses. Use these in your tooling and CI.
```
danieltprice: I wonder if a more details explanation here might be helpful. For example:

Platform API keys are bearer tokens for the Neon management API. Use them to manage your setup programmatically: create projects and branches, adjust project and compute settings such as autoscaling and IP Allow, query usage metrics, and manage roles. They come in personal, organization, and project-scoped variants, and are meant for your tooling and CI.

content/docs/concepts/credentials-access.md
```diff
@@ -0,0 +1,138 @@
...
+- **Service credentials** do your running app's work, like an object read or a model call. Use these in your app's runtime environment.
```
danieltprice: more detailed description?

Service credentials are branch-scoped bearer tokens that your app uses at runtime to call Neon's  services, such as AI Gateway and Object Storage. A single credential API mints them, and each one is limited to a specific scope (for example, ai_gateway:invoke).

content/docs/concepts/credentials-access.md
```diff
@@ -0,0 +1,138 @@
...
+
+![Diagram of Neon's two credential kinds: platform API keys that manage your Neon setup, and service credentials your running app uses, with a platform API key issuing service credentials in one direction only.](/docs/concepts/credentials-access.png 'no-border')
```
danieltprice:
```suggestion
Do we need to go through the rest of the docs and make sure we're using the "platform API keys" term consistently? For example, we don;t mention "platform API keys" on this main API page: https://neon-next-git-lkb-16595-credentials-access-neondatabase.vercel.app/docs/manage/api-keys
```

content/docs/concepts/credentials-access.md
danieltprice: We should also make sure we use the "service credential" term consistently across other docs. This might be the only place we mention this term. For example, should this term appear in the ai gateway or storage docs where we have authentication sections?

bgrenon:
Thanks for the careful read — all five are addressed. Quick rundown:

1. **"Two kinds" vs. the four-row table:** reworded the intro so it no longer says "two kinds." It now just describes the two credentials by what they do, so it doesn't clash with the table right below.

2. **More detail on the API keys bullet:** kept this one short on purpose — the fuller explanation is in the "API keys" section just below it, and the reference specifics (autoscaling, IP Allow, roles) stay linked out rather than crowding the intro.

3. **More detail on the scoped credentials bullet:** expanded it to name the services (Object Storage and the AI Gateway) and to say each credential is limited to the scopes you give it.

4. **"Platform API keys" consistency:** good point — dropped "platform" so it just reads "API keys," matching the main API keys page and the rest of the docs. Updated the diagram to match too.

5. **"Service credential" consistency:** renamed it to "scoped credential" throughout, in both the prose and the diagrams, since that's what the Object Storage and AI Gateway authentication pages already use. Making the term consistent across all the other docs is a broader cleanup we're tracking separately, rather than coining a new term just on this page.

content/docs/concepts/credentials-access.md
danieltprice: Restructured the page around one credential per job. The intro now names all four credentials up front, and the chooser table maps each row to its own on-page section, so the two-versus-four mismatch is resolved.

content/docs/concepts/credentials-access.md
danieltprice: The intro bullet was replaced in the restructure. The dedicated API keys section now carries the fuller explanation: management operations, the three key kinds, and CLI auth.

content/docs/concepts/credentials-access.md
danieltprice: Scoped credentials now have their own section covering the shared credential for Object Storage and the AI Gateway, the scopes, the branch anchor, and the S3-key and bearer-token forms.

content/docs/concepts/credentials-access.md
danieltprice: Confirmed via a cross-doc check: the term is just API keys everywhere, with no platform API keys usage.

content/docs/concepts/credentials-access.md
danieltprice: Renamed to scoped credential throughout, and a cross-doc check confirms no service credential remains on this page. Aligning the term on the AI Gateway and Storage auth pages is tracked separately.

danieltprice (approved):

PR #5677, opened 2026-08-30
docs: reposition per-branch storage as unlimited on paid plans

bgrenon:
## What this changes

Neon's docs still said branch storage was capped at ~16 TB. Nothing actually breaks at that size; it's a soft limit for cost safety, not a hard wall. This PR repositions storage so the main message across the docs is: on paid plans there's no hard per-branch size limit, and storage grows with your usage.

## Why

The old "up to 16 TB" wording made Neon look less scalable than it is, and LLMs/agents kept repeating it. Leading with "no hard limit" gives agents the accurate line to quote.

## Approach

- "No hard per-branch size limit" is now the headline on all 8 storage mentions (summaries and body).
- The 16 TB soft limit and the write-throttling detail appear in exactly one place, the "What is the maximum data size?" FAQ, so we don't repeat the old number everywhere.
- Free plan (0.5 GB per project) is unchanged.
- We avoid the word "unlimited" to steer clear of an absolute claim.

## Pages changed

plans, cost-optimization, branches, projects, consumption-limits, legacy-plans, the Neon-vs-Supabase comparison, and public/pricing.md.

## Follow-up (separate)

The vendored, read-only pricing-and-plan-features.md (mirrored from neon-for-agent-platforms) still says "Up to 16 TB" and needs a separate upstream PR.


content/changelog/2026-08-30.md
```diff
@@ -0,0 +1,11 @@
+---
```
danieltprice: Not sure we need to update an already published changelog here.

danieltprice:
So that users don't hit the soft limit and wonder what's up, perhaps we can message something like this:
"Neon has no hard storage size limit. There is only a soft limit (16 TB) to help customers avoid unexpected costs. If you require more than 16 TB of storage, please reach out to Neon Support.

danieltprice (approved):

PR #5707, opened 2026-09-02
fix: run OG image routes on the Node.js runtime

rishi-raj-jain:
Removes the `runtime = 'edge'` and `preferredRegion = 'auto'` exports from both OG image routes so they run on the default Node.js runtime instead of Edge as per the official Next.js recommendation: https://nextjs.org/docs/app/api-reference/adapters/invoking-entrypoints#edge-runtime-runtime-edge-deprecated.

Screenshot 2026-09-02 at 5 22 12 PM

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -15,6 +15,22 @@ const defaultConfig = {
   },
   outputFileTracingIncludes: {
     '*': ['./public/**/*.svg', './public/**/*.md'],
+    // OG image routes run on the Node.js runtime and read these assets from disk at
+    // request time, so they must be traced into the serverless bundle. The fonts live
+    // under src/ (not traced by default) and the images are otherwise dropped by the
+    // ./public/** exclude above.
+    '/api/og': [
+      './src/fonts/esbuild/ESBuild-Medium.ttf',
+      './src/fonts/inter/Inter-Regular.ttf',
+      './public/images/og-image/logo.png',
+      './public/images/og-image/background.png',
+    ],
+    '/docs/og': [
+      './src/fonts/geist-mono/GeistMono-Regular.ttf',
+      './src/fonts/inter/Inter-Regular.ttf',
+      './public/images/og-image/logo.png',
+      './public/images/og-image/docs-background.jpg',
+    ],
   },
   images: {
     formats: ['image/avif', 'image/webp'],
--- a/src/app/(docs)/docs/og/route.jsx
+++ b/src/app/(docs)/docs/og/route.jsx
@@ -1,8 +1,5 @@
 import { ImageResponse } from 'next/og';
 
-export const runtime = 'edge';
-export const preferredRegion = 'auto';
-
 export async function GET(request) {
   const fontBreadcrumbs = fetch(
     new URL('../../../../fonts/geist-mono/GeistMono-Regular.ttf', import.meta.url)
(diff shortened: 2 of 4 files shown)


philip:
Claude here (reviewing at Philip's request after we tested) 👋

Removing the `edge`/`preferredRegion` exports clears all three deprecation warnings, so the direction is right. But the runtime switch breaks OG image generation — worth fixing before merge.

**What I observed** (reproducible):

| | `/api/og?title=…` |
|---|---|
| This PR's preview | **500**, empty body |
| Another branch's preview | 200, `image/png`, 70899 bytes |
| Production | 200, `image/png`, 70899 bytes |

The 500 is isolated to this branch, so it's the Node runtime move, not the preview env.

**Likely cause** (inferred — I didn't pull the runtime logs): both routes load fonts/images with `fetch(new URL('...', import.meta.url))`, which resolves on Edge but throws on Node. Those fetches sit *before* the `try/catch` in `src/app/api/og/route.jsx`, which matches the empty-body 500.

**Suggested fix:** read the assets from disk in both OG routes:

```js
import { readFile } from 'node:fs/promises';
import { join } from 'node:path';

const fontTitle = await readFile(join(process.cwd(), 'src/fonts/esbuild/ESBuild-Medium.ttf'));
```

(same for the other fonts and the background images).

Repro:
```
curl -s -o /dev/null -w "%{http_code} %{content_type}\n" \
  "https://<preview>/api/og?title=$(printf 'test' | base64)"
```

rishi-raj-jain:
Appreciate the review, @philip and Claude.

I made the assumption that we'll test this directly on preview deployment and that did reveal the issue underneath so that's great.

I've simply referenced the fonts in the runtime to be included (as it'd automatically on basis of import.meta.url at build time IIRC).

I tested my changes that I've pushed locally to ensure that it does indeed work with npm run dev, npm run build && npm run start.

| URL | Preview |
| --- | --- |
| [on Localhost og image](http://localhost:3000/api/og?title=Q2xlYW4gdXAgb3JwaGFuZWQgUzMgb2JqZWN0cyB3aXRoIE5lb24gT2JqZWN0IFN0b3JhZ2UgYnJhbmNoaW5n) | image |
| [on Production og image](https://neon.com/api/og?title=Q2xlYW4gdXAgb3JwaGFuZWQgUzMgb2JqZWN0cyB3aXRoIE5lb24gT2JqZWN0IFN0b3JhZ2UgYnJhbmNoaW5n) | image |

Let's deploy to preview to validate the changes?

philip (approved):
Claude again (reviewing at Philip's request after we re-tested the preview) 👋

Confirmed both OG routes render cleanly now (`/api/og` and `/docs/og` both return valid PNGs) and the deprecation exports are gone. Nice catch wiring up `outputFileTracingIncludes` for the route-group path too.

Two optional refinements, both **pre-existing** (the old code had them too, so not blocking) — but since the rewrite already touches this spot, it could be a natural moment to fold them in:

1. **Hoist the asset loads to module scope.** They currently run on every request in both old and new code; reading the fonts/images once per instance would trim per-request work:
   ```js
   const assets = Promise.all([readFile(...), readFile(...), ...]); // module scope
   export async function GET(request) {
     const [fontA, fontB, logoBuf, bgBuf] = await assets;
   ```
2. **Bring the reads inside the `try`.** They sit just before it, so a missing asset surfaces as an empty 500 rather than the friendly `Failed to generate the image` response.

PR approved — let me know if you'll address the above or leave them for a follow-up.

rishi-raj-jain:
Appreciate the review, @philip and Claude.

Shipped 🫡

philip:
Great, thanks @rishi-raj-jain ... merging!

PR #5655, opened 2026-08-27
Concepts: add the object model page and a Concepts nav lane (LKB-16593)

bgrenon:
## Summary

Introduces a top-level **Concepts** node in the navigation, and rewrites the object-model page so a Neon branch reads as the whole backend, not just a Postgres database. This is the first page for this section; later pages coming in separate PR.

## Preview
https://neon-next-git-lkb-16593-concepts-object-model-neondatabase.vercel.app/docs/concepts/the-object-model

## Planned Concepts section

```
Concepts
├─ Backend primitives
│   ├─ Overview                 (planned)
│   ├─ The object model       ← this PR (LKB-16593)
│   └─ Branch your backend      (planned · LKB-16594)
└─ Backend behavior
    ├─ Credentials & access     (planned · LKB-16595)
    ├─ Regions & availability   (planned · LKB-16600)
    └─ Serverless & scaling     (planned · LKB-16600)
```

Tracked under epic LKB-16591; later Concepts PRs stack on this branch.

LKB-16593

PR #5771, opened 2026-09-07
docs: Branch your backend concept hub (LKB-16594)

bgrenon:
## Branch your backend

A new Concepts page on how branching works across the whole Neon backend. One branch action forks whichever services are enabled on the parent — Lakebase Postgres, Managed Better Auth, Object Storage, Functions, and the AI Gateway — each in its own way, and the page is explicit about what a restore does and does not roll back.

### Preview
https://neon-next-git-lkb-16594-branch-your-backend-neondatabase.vercel.app/docs/concepts/branch-your-backend

### Stacked on #5655
Based on the object model PR (#5655), which adds the Concepts section this page lives under. Please land #5655 first; this PR retargets to `main` automatically once it does.

### Included
- The page, with a hero diagram and a per-service diagram for each product's branching behavior.
- A 301 redirect from the retired `guides/branching-intro`, plus a few incidental link updates.


content/docs/concepts/branch-your-backend.md
```diff
@@ -0,0 +1,150 @@
...
+  <div style={{ flex: '1 1 50%' }}>
+
+![Each branch has its own AI Gateway endpoint and credentials, but shares one global model catalog.](/docs/concepts/branch-ai-gateway.png 'no-border')
+
+  </div>
+</div>
+
+Branching does not change the shape of your account or project. For what contains what, see [The Neon object model](/docs/concepts/the-object-model).
+
+## Why branching is instant
```
danieltprice, line 113: This seems to only cover Postgres branching. Maybe revise it to be inclusive of the other services and how they are also instant.

content/docs/concepts/branch-your-backend.md
```diff
@@ -0,0 +1,150 @@
...
+
+You can create a branch from the parent's current state, or from an earlier point within the project's [history window](/docs/introduction/history-window). To pick up later changes from the parent, [reset the branch from its parent](/docs/guides/reset-from-parent).
+
+This works because storage is separate from compute: durable data can start a new branch without being copied first, and each branch runs its own compute against that data. See [The lakebase architecture](/docs/introduction/architecture-overview).
+
+## Limits and caveats
+
+- **A restore rolls back Postgres and Auth only.** Restoring a branch rolls back its Postgres timeline, including Managed Better Auth data stored in the `neon_auth` schema. It does not roll back Object Storage buckets or objects, and it does not roll back deployed functions. See [Backup & restore](/docs/guides/backup-restore).
+- **AI Gateway configuration is global.** Branches get their own AI Gateway endpoint, credentials, and usage metering, but the model catalog, routing, and rate limits are shared. A branch can call a different model by sending a different `model` value in the request; it cannot have its own model catalog.
```
danieltprice, line 128: We say config is global but then we say each branch gets its own endpoint, which is a little hard to grok at first. What I'd also like to know is if AI Gateway usage can be tracked/viewed or managed per branch.

danieltprice (approved):
A couple of comments.

content/docs/concepts/branch-your-backend.md
bgrenon: Good question — reworded the section to separate the two ideas: per branch you get your own endpoint and credentials, and usage is metered per branch; the model catalog, routing, and rate limits are set at the account level and shared by every branch.

On viewing/managing usage per branch specifically: usage is metered per branch, but there is no per-branch usage view in the console or API today (that is on the roadmap), and rate limits and spend caps are account-level, not per branch. The page avoids implying a per-branch view exists.

content/docs/concepts/branch-your-backend.md
bgrenon: Good call — revised "Why branching is instant" to lead with all services (the whole backend forks in seconds), using Lakebase Postgres as the concrete copy-on-write example and noting the others follow the same principle by inheriting state or getting their own branch-scoped endpoints.

PR #5783, opened 2026-09-08
docs: add guide for migrating SQLite to Neon Postgres with AI agents and MCP

rishi-raj-jain:
Brief: /dashboard/neon/actions/sqlite-to-neon-agent-cookbook

A guide that covers how to migrate a SQLite database to Neon Postgres with an AI coding agent driving the Neon MCP Server. Covers both pgloader v4 (latest, Java based) and previous versions (Lisp) variations

### DIFF
--- a/content/guides/migrate-sqlite-neon-mcp.md
+++ b/content/guides/migrate-sqlite-neon-mcp.md
@@ -1,0 +1,330 @@
+---
+title: 'Migrate from SQLite to Neon Postgres with AI coding agents and the Neon MCP Server'
+subtitle: 'Drive the migration from an AI coding agent like Claude Code, Cursor, or Codex: provision, load with pgloader, and verify through prompts.'
+author: rishi-raj-jain
+enableTableOfContents: true
+createdAt: '2026-09-08T00:00:00.000Z'
+updatedOn: '2026-09-08T14:32:21.747Z'
+---
+
+Learn how to migrate a SQLite database to Neon Postgres with an AI coding agent through the [Neon MCP Server](https://github.com/neondatabase/mcp-server-neon). The agent creates the project, returns a connection string, generates the [pgloader](https://pgloader.readthedocs.io/en/latest/intro.html) load file, runs the migration, and checks the result.
+
+The agent uses MCP tool calls to provision and inspect the database, and spawns a terminal to run pgloader for the data transfer. pgloader streams rows over a direct Postgres connection and applies the required type casts for the SQLite migration.
+
+<Admonition type="important" title="Security">
+The Neon MCP Server can create and delete Neon resources. Review and authorize each action the agent proposes before it runs. Learn more in our [MCP security guidance](/docs/ai/neon-mcp-server#mcp-security-guidance).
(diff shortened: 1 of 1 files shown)


danieltprice:
## Live test results

I ran the full workflow end to end against a live throwaway Neon project (Postgres 18): drove the MCP steps through the Neon MCP Server, then ran pgloader for **both** documented builds. Everything works.

| # | Check | Path | Result |
|---|-------|------|--------|
| 1 | `seed.sql` builds the SQLite DB (3 authors, 5 books) | local | ✅ |
| 2 | `create_project` / `get_connection_string` via MCP | MCP | ✅ |
| 3 | Lisp: endpoint-in-password string loads | pgloader 3.6.10 | ✅ 8 rows, 0 errors |
| 4 | v4: plain direct string loads | pgloader v4.0.0 (JDBC) | ✅ 8 rows, 0 errors |
| 5 | Endpoint-in-password fails on v4 (SNI) | v4 JDBC | ✅ fails as documented |
| 6 | Type casts: `id` int+seq default, `date`, `timestamptz`, `boolean`, `jsonb`, `price` numeric uncast | MCP | ✅ exact match, both builds |
| 7 | Data integrity (0/1 → bool, JSON → jsonb queryable, timestamps) | MCP | ✅ |
| 8 | Sequence reset past max id (`last_value` 5, `nextval` 6) | MCP | ✅ |

All code samples and connection-string transformations execute exactly as written. A few notes for your consideration, all in the **pgloader v4 (JDBC)** tab:

**Accuracy:**
- **"runs on any machine with a JVM"** — the v4 jar needs Java 21+. On Java 17 it throws `NoClassDefFoundError: java/util/SequencedCollection`. Suggest "runs on any machine with Java 21 or later."
- **"It reports `pgloader v4`"** — the `v4-dev` jar actually prints `pgloader v4.0.0`.
- v4 is currently only published as a `v4-dev` **pre-release** (latest stable is 3.6.9). It's presented here as the co-equal first tab with no note that it's a development build. Worth a "pre-release" caveat, or leading with the stable Lisp build.
- The Install section only covers the Lisp build (brew/apt); the v4 tab never says where `pgloader.jar` comes from. A download line would help (it's on the `v4-dev` release page).

**Endpoint-in-password failure, precise behavior:** the note says moving the endpoint into the password on v4 fails with `password authentication failed`. Exactly:
- raw `;` → **parse error** at the `;` (the load-file parser breaks first)
- `%3B`-encoded `;` → `password authentication failed` (as stated)

Since the JDBC tab tells readers to leave the endpoint out, neither failure is hit in practice. Optional to mention.

**Minor wording:** "The easiest way to setup..." → "set up" (verb); "PostgreSQL JDBC driver" reads inconsistently with "Its Postgres driver" in the Lisp tab.

Tested 2026-09-11. Great guide overall, the type-cast coverage and the SNI explanation are spot on.

rishi-raj-jain:
Thanks for the quick review! I've made the change from "setup" to "set up". Otherwise, I believe even AI liked my work! This guide is made FOR agents actually 😉

PR #5790, opened 2026-09-08
docs: add Function Triggers guides for schedule and object uploads

philip:
## Overview

Add documentation for Neon Function Triggers: an overview of the feature and how triggers behave across branches, plus a worked guide for each trigger type, `schedule` (cron) and `storage_object_created` (object uploads).

Each guide shows how to create and manage triggers from the Neon Console, the `neon triggers` CLI, the Neon API, and (for schedule) `neon.ts`, with diagrams and the request envelope a trigger delivers. Examples were verified end to end against a live branch.

### DIFF
--- a/content/docs/compute/functions/triggers/overview.md
+++ b/content/docs/compute/functions/triggers/overview.md
@@ -1,0 +1,93 @@
+---
+title: Function Triggers
+subtitle: Let Neon invoke a function for you.
+summary: >-
+  Function Triggers invoke a deployed Neon Function on a cron schedule, with no external
+  scheduler and no compute kept running to hold the timer. Covers what a scheduled invocation
+  sends your function, how triggers behave across branches, and the current limits. Schedule
+  is the only trigger type today, evaluated in UTC.
+enableTableOfContents: true
+updatedOn: '2026-09-08T22:56:13.993Z'
+---
+
+A Function Trigger tells Neon to invoke a deployed [Neon Function](/docs/compute/functions/overview) on a schedule. You give it a cron expression, and Neon calls the function at each run. No external scheduler, and no compute kept running to hold the timer.
+
+Today the only trigger type is `schedule`, a cron expression evaluated in UTC. The API uses a `type` discriminator, so more types can be added later without changing existing triggers.
+
+Triggers are managed through the Neon API; there are no `neon` CLI commands for them yet. To create one, see [Schedule a function](/docs/compute/functions/triggers/schedule).
+
+## Function Triggers vs pg_cron
+
+[pg_cron](/docs/extensions/pg_cron) schedules SQL inside Postgres. Function Triggers schedule your function code. They solve different problems:
+
+|                        | pg_cron                              | Function Triggers                                                                            |
+| ---------------------- | ------------------------------------ | -------------------------------------------------------------------------------------------- |
+| Runs                   | A SQL statement or Postgres function | Your JavaScript or TypeScript function                                                       |
+| Where                  | Inside the Postgres compute          | On Neon's compute, next to your data                                                         |
+| External APIs          | No                                   | Yes: HTTP, [AI Gateway](/docs/ai-gateway/overview), [Object Storage](/docs/storage/overview) |
+| Compute scaled to zero | Doesn't run                          | Runs; the invocation starts the function                                                     |
+
+## Trigger fields
+
+A trigger belongs to a project and branch and points to one function on that branch.
+
+| Field           | Required | Description                                                                                           |
+| --------------- | -------- | ----------------------------------------------------------------------------------------------------- |
+| `type`          | Yes      | `"schedule"` is the only value today.                                                                 |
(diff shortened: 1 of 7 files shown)


content/docs/compute/functions/triggers/overview.md
```diff
@@ -0,0 +1,93 @@
...
+  scheduler and no compute kept running to hold the timer. Covers what a scheduled invocation
+  sends your function, how triggers behave across branches, and the current limits. Schedule
+  is the only trigger type today, evaluated in UTC.
+enableTableOfContents: true
+updatedOn: '2026-09-09T05:14:01.852Z'
+---
+
+A Function Trigger tells Neon to invoke a deployed [Neon Function](/docs/compute/functions/overview) on a schedule. You give it a cron expression, and Neon calls the function at each run. No external scheduler, and no compute kept running to hold the timer.
+
+Today the only trigger type is `schedule`, a cron expression evaluated in UTC. The API uses a `type` discriminator, so more types can be added later without changing existing triggers.
```
danieltprice:
```suggestion
Today the only trigger type is `schedule`, a cron expression evaluated in Coordinated Universal Time (UTC). The API uses a `type` discriminator, so more types can be added later without changing existing triggers.
```

content/docs/compute/functions/triggers/overview.md
```diff
@@ -0,0 +1,93 @@
...
+
+You manage triggers through the Neon API. To create one, see [Schedule a function](/docs/compute/functions/triggers/schedule).
+
```
danieltprice, line 21: Content looks solid, but I wonder if it might benefit from more information about when a developer might reach for this feature. Maybe add a section something like this?

## When to use a scheduled function

A scheduled function runs any recurring job you'd otherwise put in a crontab: email a morning digest, generate a weekly report, retry failed webhooks, warm a cache, or ping your own health endpoint. You write the job as a function, and Neon invokes it on the schedule.

It's especially useful when the job needs application code, an external service, or logic that's awkward to express in SQL alone:

- **Roll up data for a dashboard.** Aggregate the day's rows into a summary table each night, so the app reads one pre-computed row instead of scanning raw events on every page load.
- **Refresh derived data from an external API.** Pull exchange rates, inventory, or embeddings from the [AI Gateway](/docs/ai-gateway/overview), then upsert the results into Postgres.
- **Expire and clean up.** Delete rows past their TTL, close abandoned carts, or archive stale records on a fixed cadence, next to the data instead of over the public internet.
- **Sync from a third-party system.** Import a payments or CRM export every few minutes so the rest of your app queries local Postgres, not a rate-limited API.

These jobs also vary in how long they take, and that's fine here. Most serverless platforms cap how long a function can run, so a heavy nightly batch has to be broken up or moved to a separate service. Neon Functions are long-running, so the same scheduled function handles both a quick ping and a table-scanning batch, with nothing extra to configure for the larger job.

content/docs/compute/functions/triggers/schedule.md
```diff
@@ -0,0 +1,273 @@
...
+title: Schedule a function
+subtitle: Create and manage scheduled Function Triggers.
+summary: >-
+  Create, list, update, disable, and delete scheduled Function Triggers with the Neon API:
+  a Hono handler for the scheduled POST, a five-field UTC cron reference, how to confirm a
+  run in the logs, and the common errors.
+enableTableOfContents: true
+updatedOn: '2026-09-09T05:14:01.852Z'
+---
+
```
danieltprice, line 11: This might be a good opportunity to add a prompt in some form, maybe a copy prompt button or natural language prompt?

```
Build and schedule a Neon Function.

- Expose a single POST route at "/" with no auth on that route: scheduled
  invocations arrive without credentials. Use whatever HTTP framework the project
  already uses (Hono, Express, Fastify, or plain Node). Read `scheduled_at` from
  the JSON body and keep the handler idempotent so a redelivered run is safe.
- Connect to Postgres with the injected DATABASE_URL (any client works;
  @neondatabase/serverless is a good default).
- The job should: <describe what it does>.

Deploy the function, then create a schedule trigger with the Neon API. Cron is
five numeric UTC fields (no macros, named days, or seconds). Start the cron at
`* * * * *` to confirm a run in the logs, then PATCH it to the real cadence.

I loosened the DB client line for the same reason, so the agent uses pg, postgres.js, Prisma, or Drizzle if that's what the project already has, rather than forcing the Neon driver.

Prose version:

Create a Neon Function that does <task>, then schedule it. Add one unauthenticated POST route to the app using whatever HTTP framework the project already uses; the route reads `scheduled_at` from the body and is safe to run twice. Connect to Postgres with the injected DATABASE_URL. Deploy it, then create a schedule trigger via the Neon API using a five-field UTC cron, starting at `* * * * *` to verify a run before switching to the real cadence.
```

danieltprice:
Heads-up for when this lands: a Console UI for schedule trigger management was added in `databricks-eng/universe#2553038` (currently gated, not yet live). Once it ships, the "managed through the Neon API" wording on the overview and schedule pages may want to acknowledge the Console too. API-first is fine for launch — just flagging so the docs don't stay API-only by default.

PR #5829, opened 2026-09-11
Add LibreDB Studio to the tested GUI applications table

yusuf-gundogdu:
Adds LibreDB Studio to the tested GUI applications and IDEs table.

LibreDB Studio is an MIT licensed, self hosted SQL IDE that runs in the browser. You deploy one container next to your databases and open a URL instead of installing a client on every machine.

Tested against a Neon Free project on 11 September 2026, PostgreSQL 18.6, using the 0.15.0 Docker image. Pasting the Neon connection string into the connection form filled in the host, role, password and database, and set SSL to require on its own because of sslmode=require in the string. The connection came up in 832 ms with no SNI workaround and no manual changes. The object browser listed both test tables with correct row counts, a join and group by query returned in 331 ms, and EXPLAIN rendered the full plan tree.

Nothing needed a workaround, so the notes cell is left empty.

Home page: https://libredb.org
Repo: https://github.com/libredb/libredb-studio


yusuf-gundogdu:
Checking in on this one.

The branch is still clean against main and the only red check is the Vercel preview, which is waiting on a team member to authorize a deployment from a fork rather than on anything in the change itself. The table entry is a single row and I am happy to reword it if it should read differently.

PR #5879, opened 2026-09-17
docs: backend GA follow-ups (credit purchases, settings page)

danieltprice:
- AI Gateway prepaid credits: note that a declined card fails the purchase, adds no credits, leaves no payable invoice, and doesn't appear in credit history.
- Project settings: update the Settings section for the redesigned tabbed page (General, Postgres, Networking, Credentials, Auth, Functions, HIPAA compliance, Sharing). Remove dated screenshots; text only.
- Object Storage overview: drop the "when billing begins" hedge now that the backend is GA; keep object size consistent with the limits table.
- Next.js FAQ: drop the stale "Functions, Object Storage, and AI Gateway are in beta" note.


philip (approved):
I wavered on keeping `preview` (as old neon/config versions work on both) but using the latest variant makes sense too. We have a troubleshooting guide and a few strategic `preview` notes, so 👍

PR #5883, opened 2026-09-17
docs: update TypeScript SDK reference for v5/v6 named parameters

danieltprice:
Updates the TypeScript SDK reference to match `@neon/sdk` v5 and later, where the current major is v6.

## Background

Starting in v5, the SDK replaced positional arguments with a **single named-parameter object**. Path selectors (`projectId`, `branchId`, resource ids like `roleName` or `slug`) and any input or query fields are merged into one flat object, followed by the optional trailing options argument. For example:

```ts
// v4 and earlier
neon.branches.get(projectId, branchId);

// v5+ (v6 current)
neon.branches.get({ projectId, branchId });
```

The reference page still documented the old positional form, so calls copied from the docs would not compile against a current install.

## Changes

- Rewrite the calling-convention overview to describe the single named-parameter object.
- Add an "Named parameters in v5 and later" admonition linking to the existing **Upgrading from v4** section.
- Update every method signature in the reference tables and all inline code examples to the object form (`get({ projectId })`, `branches.create({ projectId, name })`, `operations.waitFor({ operations })`, and so on).

Reference-only doc change. No navigation or component changes.


danieltprice:
## Live test results

Tested every documented calling convention against a real `@neon/sdk@6.0.0` install (current major; npm also has `5.0.0`, `4.2.0`, `4.1.0`, matching the version claims). Three passes: type-check every named-parameter signature/example in the doc, a negative check that the old positional forms no longer compile on v6, and a read-only run against the live Neon API.

| # | Test | Type | Result |
|---|------|------|--------|
| 1 | All ~80 named-parameter signatures + inline examples | Type check (v6) | ✅ Passed (`tsc --noEmit` exits 0) |
| 2 | Old positional forms rejected on v6 | Type check (negative) | ✅ Passed |
| 3 | `projects.list().all()` | Live | ✅ Passed |
| 4 | `projects.get({ projectId })` | Live | ✅ Passed |
| 5 | `branches.list({ projectId }).all()` | Live | ✅ Passed |
| 6 | `branches.getDefault({ projectId })` | Live | ✅ Passed |
| 7 | `branches.get({ projectId, branchId })` | Live | ✅ Passed |
| 8 | `postgres.roles.list({ projectId, branchId })` | Live | ✅ Passed |
| 9 | `postgres.databases.list({ projectId, branchId })` | Live | ✅ Passed |
| 10 | `postgres.endpoints.listByBranch({ projectId, branchId })` | Live | ✅ Passed |
| 11 | `postgres.endpoints.list({ projectId })` | Live | ✅ Passed |
| 12 | `operations.list({ projectId }).all()` | Live | ✅ Passed |
| 13 | `apiKeys.list()` / `user.me()` / `regions.list()` | Live | ⚠️ Blocked — account-level endpoints reject the org API key used for testing; not a doc/SDK issue. Shapes confirmed in test 1. |

**No doc errors found.** Every named-parameter signature and example (including field names like `roleName`, `databaseName`, `bucketName`, `objectKey`, `keyName`, `slug`) type-checks against v6, and the project/branch/Postgres/operations calls run end-to-end against the live API. The negative pass confirms the pre-PR positional examples genuinely fail to compile on the current SDK, so the update was both needed and accurate.

PR #5884, opened 2026-09-17
docs: clarify shared credentials security vs branding

danieltprice:
Follow-up to #5882. Adds a paragraph to the **About shared credentials** section answering whether shared credentials are a security/impersonation risk or just a branding matter, a question raised about the Google OAuth provider dialog in the console.

## Change

Clarifies that shared credentials are **one Neon-owned Google OAuth app shared across every project that uses them**, not an app scoped to your project, and that this is a branding/control trade-off rather than an impersonation risk.

## Verified against neon-auth source

The wording was checked against `neon-cloud/neon-auth`:

- Shared Google uses service-level env vars `SHARED_GOOGLE_OAUTH_CLIENT_ID` / `_SECRET` (`config.ts`) — one app for the whole deployment (`social-providers.ts`).
- Sign-in redirects to a single shared callback `${NEON_AUTH_SERVER_URL}/auth/oauth/callback/google`, not a per-project callback.
- Per-project isolation at that shared endpoint is enforced by HMAC-signed OAuth state carrying `endpointId` (`oauth-state.ts`, `SHARED_OAUTH_HMAC_KEY`, commented "shared across tenants").
- Supplying your own Client ID/secret switches the project to your own app instead.

Auth team: please sanity-check the security framing in the final paragraph, but the mechanics above are drawn directly from the source.


content/docs/auth/guides/setup-oauth.md
```diff
@@ -23,9 +23,13 @@ Google OAuth is enabled by default with shared credentials for development and t
...
+Shared credentials let you add Google sign-in with no setup, using a Google OAuth app that Neon owns. That one app is shared by every project that uses shared credentials, including projects belonging to other Neon users. It isn't unique to you.
 
-Use shared credentials only for development and testing. For production, add your own Google Client ID and secret so users see your app's branding and you control the credentials. See [Production setup](#production-setup).
+When someone signs in, Google shows them Neon's name and logo, not your app's, because the sign-in runs through Neon's app rather than one you own. Your users are effectively trusting Neon, and you don't control the app's branding, its permission scopes, or the credentials.
+
+This is safe for development and testing. An outside website can't use the shared app to capture your users, because Google only ever redirects sign-ins back to Neon, and Neon keeps each project's sign-ins separate. The reason not to use it in production is trust and control, not a security hole: real users should see your app on the consent screen, and you should own the credentials.
```
danieltprice:
```suggestion
This is safe for development and testing. An outside website can't use the shared app to capture your users, because Google only ever redirects sign-ins back to Neon, and Neon keeps each project's sign-ins separate. The reason not to use it in production is trust and control. Real users should see your app on the consent screen, and you should own the credentials.
```

PR #5894, opened 2026-09-18
docs: fix full-backend quickstart function for services GA

danieltprice:
Updates the AI function in the full backend quickstart after the Object Storage / Functions / AI Gateway GA. Verified with a live end-to-end run.

- Switch AI Gateway models from frontier (`gpt-5-nano`/`gpt-5-mini`) to open-weight (`gpt-oss-20b`, `meta-llama-3-3-70b-instruct`), which are enabled by default. Frontier models require requesting access, so the guide's originals returned `model requires a verified account`.
- Cast `created_at` to text in the assistant's `queryPosts` tool. A raw pg `Date` in the tool result fails the AI SDK v7 `ModelMessage[]` schema on the step after the tool runs, so the assistant never answered.


danieltprice:
## Live testing coverage

Ran the full guide end-to-end against a real Neon project (org with credits, region **AWS US East / N. Virginia** — a newly supported region) using **Neon CLI 5.0.0**. Scaffolded `create-next-app`, followed every step, and exercised each service against the live API.

| Area | Coverage | Result |
| --- | --- | --- |
| CLI flow | `neon link --no-config` / `checkout` / `config init` | ✅ |
| Data layer | `drizzle-kit push` + seed + published-only query | ✅ (draft hidden) |
| `neon.ts` GA syntax | top-level `buckets` / `functions` / `aiGateway` (no `preview:`) | ✅ deployed |
| Object Storage | Files SDK upload → public URL `GET` | ✅ 200 |
| Function deploy | bucket + function provisioned, URL + creds injected | ✅ |
| `POST /generate` | `gpt-oss-20b` → generates and saves a post | ✅ 200 |
| `POST /assistant` | `meta-llama-3-3-70b-instruct`, tool loop queries Postgres and streams answer | ✅ 0 errors |
| Client code | `tsc` against `ai@7.0.97` / `@ai-sdk/react` | ✅ clean |

### Issues this PR fixes
- **Frontier models fail.** `gpt-5-nano` / `gpt-5-mini` return `model requires a verified account` (500). Only open-weight models are enabled by default.
- **Assistant crash.** Raw pg `Date` (`created_at`) in the tool result throws `AI_InvalidPromptError: messages do not match the ModelMessage[] schema` on the post-tool step. Failed identically across three models (model-agnostic). Casting to text fixes it.

### Verified, no change needed
- Region set is now the four regions listed on the page (confirmed by CLI error/help output).
- AI Gateway still requires a paid plan (deploy hard-fails on Free).
- Object Storage injects `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_ENDPOINT_URL_S3` / `AWS_REGION`; AI Gateway injects `NEON_AI_GATEWAY_TOKEN` / `NEON_AI_GATEWAY_BASE_URL`.

### Follow-ups (not in this PR)
- **`neon deploy` still prints "AI Gateway is in public beta"** even though the `neon.ts` keys are GA — CLI/product surface out of sync with the GA docs.
- **Function URL auto-injection:** deploy now injects `NEON_FUNCTION_POSTS_BASE_URL` (no trailing slash) — the manual `NEXT_PUBLIC_POSTS_FN_URL` step still works but could be simplified.
- **`files-sdk` install hits ERESOLVE** because `create-next-app@latest` currently pulls `next@16` canary; `--legacy-peer-deps` works. Likely transient.

This comment was written by Isaac.

